PR #14, opened 2021-11-01
JoeMakerV2

LouisMeMyself:
Takes into account reflect tokens, and forking mainnet for tests


contracts/JoeMakerV2.sol
```diff
@@ -0,0 +1,231 @@
...
+        pair.burn(address(this));
+
+        // We get the amount0 and amount1 by their respective balance of JoeMakerV2.
+        uint256 amount0 = IERC20(token0).balanceOf(address(this));
+        uint256 amount1 = IERC20(token1).balanceOf(address(this));
+
+        // we invert token0 and token1 as well as we'll need them to get their respective balance.
+        if (token0 != pair.token0()) {
+            (amount0, amount1) = (amount1, amount0);
+            (token0, token1) = (token1, token0);
```
cryptofish7, line 116: What's the purpose of this line?

LouisMeMyself: It's useless, I didn't see that when I changed the way to retrieve the amounts, there was no need to exchange the tokens or the amounts.

PR #25, opened 2021-12-26
sJoe

LouisMeMyself:
sJoe is a contract that allows to deposit moJoe and receives stable coin sent by JoeMakerV4's daily harvests. Users deposit moJoe and receive a share of what has been sent by JoeMakerV4 based on their participation of the total deposited moJoe.


contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+        emit Withdraw(msg.sender, _amount);
+    }
+
+    // Withdraw without caring about rewards. EMERGENCY ONLY.
+    function emergencyWithdraw() public {
+        UserInfo storage user = userInfo[msg.sender];
+
+        uint256 _amount = user.amount;
+        user.amount = 0;
+        user.rewardDebt = 0;
```
jummy123, line 128: Is there any benefit to setting `user.rewardDebt` to 0 here? In both `deposit` and `withdraw` `rewardDebt` is set to a new value so not sure if there is some other reason to do it here?

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+// of the total deposited moJoe.
+
+contract StableJoeStaking is Initializable, OwnableUpgradeable {
+    using SafeMathUpgradeable for uint256;
+    using SafeERC20Upgradeable for IERC20Upgradeable;
+
+    // Info of each user.
+    struct UserInfo {
+        uint256 amount; // How many LP rewardTokens the user has provided.
+        uint256 rewardDebt; // Reward debt. See explanation below.
```
jummy123, line 22: Suggestion for updating this comment

> The sum of rewards already paid to this account. See explanation below

jummy123:
Should there be a method to withdraw just the accumulated balance of the reward token. Right now it looks like the only way to claim rewards is either depositing or withdrawing. I can imagine a case where I want to take the reward now but leave my moJoe staking.

jummy123:
The logic of the contract LGTM :+1:

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+
+import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/math/SafeMathUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/token/ERC20/SafeERC20Upgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/Initializable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+// StableJoeStaking is a contract that allows to deposit moJoe and receives stable coin sent by JoeMakerV4's daily
+// harvests. Users deposit moJoe and receive a share of what has been sent by JoeMakerV4 based on their participation
+// of the total deposited moJoe.
```
cryptofish7, line 13: Use natspec comments

contracts/StableJoeStaking.sol
cryptofish7, line 22: Use natspec for each of the member variables

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+    // Info.
+    IERC20Upgradeable moJoe; // Address of moJoe contract.
+    uint256 lastTokenBalance; // Last balance of reward token.
+    uint256 accTokenPerShare; // Accumulated Tokens per share, times 1e12. See above.
+
+    // The reward token.
+    IERC20Upgradeable public rewardToken;
+
+    // Info of each user that stakes moJoe.
+    mapping(address => UserInfo) public userInfo;
```
cryptofish7, line 45: Use natspec comments for each variable

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+
+    event Deposit(address indexed user, uint256 amount);
+    event Withdraw(address indexed user, uint256 amount);
+    event EmergencyWithdraw(address indexed user, uint256 amount);
```
cryptofish7, line 49: Use natspec for each event

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+
+    function initialize(IERC20Upgradeable _rewardToken, IERC20Upgradeable _moJoe) public initializer {
```
cryptofish7, line 51: Use natspec comment

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+        __Ownable_init();
+
+        rewardToken = _rewardToken;
+        moJoe = _moJoe;
+    }
+
+    // View function to see pending reward token on frontend.
+    function pendingToken(address _user) external view returns (uint256) {
```
cryptofish7, line 59: Use natspec

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+
+        uint256 rewardTokenBalance = rewardToken.balanceOf(address(this));
+        if (rewardTokenBalance != lastTokenBalance && lpSupply != 0) {
+            uint256 rewardTokenAdded = rewardTokenBalance.sub(lastTokenBalance);
+            _accTokenPerShare = _accTokenPerShare.add(rewardTokenAdded.mul(1e12).div(lpSupply));
+        }
+        return user.amount.mul(_accTokenPerShare).div(1e12).sub(user.rewardDebt);
+    }
+
+    // Update reward variables to be up-to-date.
```
cryptofish7, line 72: Use natspec comment

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+            safeTokenTransfer(msg.sender, pending);
+        }
+        user.amount = user.amount.add(_amount);
+        user.rewardDebt = user.amount.mul(accTokenPerShare).div(1e12);
+
+        moJoe.safeTransferFrom(address(msg.sender), address(this), _amount);
+        emit Deposit(msg.sender, _amount);
+    }
+
+    // Withdraw moJoe from sJoe and harvest the rewards.
```
cryptofish7, line 106: Use natspec comment

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+
+        user.amount = user.amount.sub(_amount);
+        user.rewardDebt = user.amount.mul(accTokenPerShare).div(1e12);
+
+        safeTokenTransfer(msg.sender, pending);
+        moJoe.safeTransfer(address(msg.sender), _amount);
+        emit Withdraw(msg.sender, _amount);
+    }
+
+    // Withdraw without caring about rewards. EMERGENCY ONLY.
```
cryptofish7, line 122: Use natspec comment

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,145 @@
...
+
+        uint256 _amount = user.amount;
+        user.amount = 0;
+        user.rewardDebt = 0;
+
+        moJoe.safeTransfer(address(msg.sender), _amount);
+        emit EmergencyWithdraw(msg.sender, _amount);
+    }
+
+    // Safe Token transfer function, just in case if rounding error causes pool to not have enough reward tokens.
```
cryptofish7, line 134: Use natspec comment

contracts/StableJoeStaking.sol
LouisMeMyself: This is how MasterChef does it, I guess it's because `user` amounts can be used elsewhere on another contract.
So an exploiter could emergencyWithdraw its funds and still use the other protocol like he hadn't withdrawn anything.

LouisMeMyself:
> 

To claim rewards without withdrawing anything, you call withdraw(0), this is how masterChef does it

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,181 @@
...
+    /// @notice Info.
+    IERC20Upgradeable moJoe; /// @notice Address of moJoe contract.
+    uint256 lastTokenBalance; /// @notice Last balance of reward token.
+    uint256 accTokenPerShare; /// @notice Accumulated Tokens per share, times 1e12. See above.
+
+    /// @notice The reward token.
+    IERC20Upgradeable public rewardToken;
+
+    /// @notice Info of each user that stakes moJoe.
+    mapping(address => UserInfo) public userInfo;
```
0xnivek, line 49: nit: place `public` variables above `internal` ones

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,181 @@
...
+    event EmergencyWithdraw(address indexed user, uint256 amount);
+
+    /**
+     * @notice Initialize a new StableJoeStaking contract.
+     * @dev This contract needs to receive an ERC20 `_rewardToken` in order to distribute them
+     * (with JoeMakerV4 in our case)
+     * @param _rewardToken The address of the ERC-20 reward token
+     * @param _moJoe The address of the ERC-20 moJOE token, traderjoe's staking token.
+     */
+    function initialize(IERC20Upgradeable _rewardToken, IERC20Upgradeable _moJoe) public initializer {
```
0xnivek, line 67: nit: mark as `external`, unless theres a reason `initialize` needs to be `public`

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,181 @@
...
+        UserInfo storage user = userInfo[_user];
+        uint256 lpSupply = moJoe.balanceOf(address(this));
+        uint256 _accTokenPerShare = accTokenPerShare;
+
+        uint256 rewardTokenBalance = rewardToken.balanceOf(address(this));
+        if (rewardTokenBalance != lastTokenBalance && lpSupply != 0) {
+            uint256 rewardTokenAdded = rewardTokenBalance.sub(lastTokenBalance);
+            _accTokenPerShare = _accTokenPerShare.add(rewardTokenAdded.mul(1e12).div(lpSupply));
+        }
+        return user.amount.mul(_accTokenPerShare).div(1e12).sub(user.rewardDebt);
```
0xnivek, line 89: nit: suggest to put `1e12` into a `public constant` and used that instead since its used multiple times in this contract

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /**
+     * @notice Deposit moJoe to sJoe for reward token allocation.
+     * @param _amount The amount of moJOE to deposit
+     */
+    function deposit(uint256 _amount) public {
+        UserInfo storage user = userInfo[msg.sender];
+
+        updatePool();
```
0xnivek, line 120: nit: move `updatePool()` call to be first line of this function to make it explicit it needs to be called before proceeding with any logic

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,181 @@
...
+        if (user.amount > 0) {
+            uint256 pending = user.amount.mul(accTokenPerShare).div(1e12).sub(user.rewardDebt);
```
0xnivek, line 122: might be worth creating a `getPendingReward` function since this logic is also repeated in line 89

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,181 @@
...
+            safeTokenTransfer(msg.sender, pending);
+        }
+        user.amount = user.amount.add(_amount);
+        user.rewardDebt = user.amount.mul(accTokenPerShare).div(1e12);
+
+        moJoe.safeTransferFrom(address(msg.sender), address(this), _amount);
```
0xnivek, line 128: nit: i don't think you need the `address` around `address(msg.sender)` since `msg.sender` should already be an address I believe

0xnivek (commented):
Very new and interesting stuff I haven't seen before! Just left some minor suggestions

contracts/StableJoeStaking.sol
LouisMeMyself: I don't find a good way to do this tbh, as one use the variable in memory, the other a copy.

contracts/JoeBarV2.sol
```diff
@@ -0,0 +1,94 @@
...
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+
+interface IBarRewarder {
+    function claimReward() external;
+}
+
+// JoeBar is the coolest bar in town. You come in with some Joe, and leave with more! The longer you stay, the more Joe you get.
+//
+// This contract handles swapping to and from moJoe, JoeSwap's staking token.
+contract JoeBarV2 is Initializable, ERC20Upgradeable, OwnableUpgradeable {
```
0xnivek, line 19: nit: use natspec format

contracts/JoeBarV2.sol
```diff
@@ -0,0 +1,94 @@
...
+        uint256 totalJoe = joe.balanceOf(address(this));
+        // Gets the amount of moJoe in existence
+        uint256 totalShares = totalSupply();
+        // If no moJoe exists, mint it 1:1 to the amount (minus the fees) put in
+        if (totalShares == 0 || totalJoe == 0) {
+            require(_amount >= 1e18, "JoeBarV2: You need to enter with at least 1 $JOE.");
+            _mint(msg.sender, amount);
+        }
+        // Calculate and mint the amount of moJoe the Joe is worth. The ratio will change overtime, as moJoe is
+        // burned/minted and Joe deposited + gained from fees / withdrawn.
```
0xnivek, line 57: can we explain how the ratio is calculated here

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,254 @@
...
+
+            _tokenReward = _multiplier.mul(_tokenPerSecScaled);
+
+            _lastRewardTimestamp = _lastDailyFeeTimestamp + 1 days;
+            _lastDailyFeeTimestamp = (block.timestamp / 1 days) * 1 days;
+
+            uint256 _accruedFee = rewardToken.balanceOf(address(this)).sub(_lastTokenBalance);
+            _lastTokenBalance = _lastTokenBalance.add(_accruedFee);
+
+            _tokenPerSecScaled = _accruedFee.mul(PRECISION).div(1 days);
```
0xnivek, line 141: it'd be great to add some comments in line here explaining the math to be explicit

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,254 @@
...
+        }
+
+        if (block.timestamp > lastRewardTimestamp + 1 days) {
```
0xnivek, line 144: question: should this be `else if` or na? asking b/c it might be possible that both `block.timestamp > _lastDailyFeeTimestamp + 1 days` and `block.timestamp > lastRewardTimestamp + 1 days` is true, in which case `_multiplier` would be assigned on both line 131 and then overwritten on line 145. not sure what the expected behavior is here tho

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,254 @@
...
+     * @param _amount The amount to send to `_to`
+     */
+    function safeTokenTransfer(address _to, uint256 _amount) internal {
+        uint256 rewardTokenBal = rewardToken.balanceOf(address(this));
+        if (_amount > rewardTokenBal) {
+            rewardToken.transfer(_to, rewardTokenBal);
+            lastTokenBalance = lastTokenBalance.sub(rewardTokenBal);
+        } else {
+            rewardToken.transfer(_to, _amount);
+            lastTokenBalance = lastTokenBalance.sub(_amount);
```
0xnivek, line 251: nit: follow `checks-effects-interactions` pattern and update `lastTokenBalance` before doing the transfers

contracts/rewarders/BarRewarderPerSec.sol
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/**
+ * This is a sample contract to be used in the JoeBarV2 to reward moJoe holder with JOE.
+ *
+ * It assumes no minting rights, so requires a set amount of YOUR_TOKEN to be transferred to this contract prior.
+ * E.g. say you've allocated 100,000 XYZ to the JOE-XYZ farm over 30 days. Then you would need to transfer
+ * 100,000 XYZ and set the block reward accordingly so it's fully distributed after 30 days.
+ *
+ */
```
0xnivek, line 26: nit: natspec

contracts/rewarders/BarRewarderPerSec.sol
```diff
@@ -0,0 +1,119 @@
...
+
+        joe = _joe;
+        bar = _bar;
+        apr = _apr;
+        lastRewardTimestamp = block.timestamp;
+        tokenPerSec = 0;
+    }
+
+    /// @notice Update reward variables.
+    function update() public {
```
0xnivek, line 66: nit: lets order functions as `external -> public -> internal -> private`

0xnivek (changes requested):
great work!! 💪  left some small suggestions

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,254 @@
...
+     */
+    function initialize(IERC20Upgradeable _rewardToken, IERC20Upgradeable _joe) external initializer {
+        __Ownable_init();
+
+        rewardToken = _rewardToken;
+        joe = _joe;
+
+        lastTokenBalance = _rewardToken.balanceOf(address(this)); /// @dev Added to be upgrade safe
+        PRECISION = 1e12; /// @dev initialized here to be upgrade safe
+        lastDailyFeeTimestamp = (block.timestamp / 1 days) * 1 days;
```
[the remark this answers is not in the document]
LouisMeMyself: Solidity doesn't have any floats, so `block.timestamp / 86400 * 86400` returns the timestamp of the day at 00:00:00
(basically it's doing `int(block.timestamp / 86400) * 86400`)

contracts/StableJoeStaking.sol
LouisMeMyself: It is expected to be an `if`:
L134 is to calculate the rewards of the last day before upgrading to the new values of today's buyback
L145 is to calculate the rewards of the new day

contracts/StableJoeStaking.sol
LouisMeMyself: Indeed, I'll try to be more explicit

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,270 @@
...
+        if (block.timestamp > lastDailyFeeTimestamp + 1 days) {
+            // remaining time of the previous day
+            multiplier = 1 days - (lastRewardTimestamp % 1 days);
+
+            tokenReward = multiplier.mul(tokensPerSecScaled);
+
+            // now we update the values for the new day
+            lastRewardTimestamp = lastDailyFeeTimestamp + 1 days;
+            // get the current day at 00:00:00, _lastRewardTimestamp and _lastDailyFeeTimestamp
+            // can be different, if `updatePool` isn't called at least once per day.
```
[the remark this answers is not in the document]
LouisMeMyself: I'm trying to explain that `_lastRewardTimestamp ` and `_lastDailyFeeTimestamp` are equal unless that function isn't called for more than a day, and if so, they need to be different for calculating the rewards accordingly

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,270 @@
...
+    /// @notice The reward token sent by JoeMakerV4
+    IERC20Upgradeable public rewardToken;
+    uint256 public lastRewardTimestamp;
+
+    uint256 private tokensPerSecScaled;
+    /// @dev We update tokenPerSec after 1 day after lastDailyFeeTimestamp
+    uint256 public lastDailyFeeTimestamp;
+    /// @notice Last balance of reward token
+    uint256 public lastTokenBalance;
+    uint256 public reserves;
```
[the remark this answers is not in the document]
LouisMeMyself: It's not used at all in fact x)

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,270 @@
...
+
+            // get today's buyback amount
+            uint256 accruedFee = rewardToken.balanceOf(address(this)).sub(lastTokenBalance);
+            lastTokenBalance = lastTokenBalance.add(accruedFee);
+
+            // distribute today's buyback over 1 day
+            tokensPerSecScaled = accruedFee.mul(PRECISION).div(1 days);
+        }
+
+        // in case `updatePool` isn't called in more than 2 days, we need to make sure user receives only one day
```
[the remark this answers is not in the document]
LouisMeMyself: In fact it's not called during one full day, so
if it's called twice, one time at day 0, 23:59:59 and the second time at day 2 00:00:01 (no call during 24:00:02), it will be true
but if it's called twice, one time at day 0, 00:00:01 and the second time at day 1 23:59:59 (no call during 47:59:58), it will be true

Updated comments to `In case `updatePool` isn't called at least once per day`

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,270 @@
...
+        // worth of token
+        if (block.timestamp > lastRewardTimestamp + 1 days) {
+            tokenReward = tokenReward.add(uint256(1 days).mul(tokensPerSecScaled));
+            tokensPerSecScaled = 0;
```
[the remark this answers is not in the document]
LouisMeMyself: As we assume that a buyback happens every day, and therefore that updatePool is called at least once a day. This will never be true.

But if for what ever reason, `updatePool` isn't called for one full day (see coments above), then we still need to distribute the rewards that were accrued, but it only takes into account one day of reward.
So if we're not updating this to 0, the previous rewards will keep being distributed and we will be in debt.
After that if a buyback happen, it will be distributed the next day.

So:
day 0:
tokenPerSec = 0
accruedRewards = 86400

day 1:
tokenPerSec = 1
accruedRewards = 0

day 2:
tokenPerSec = 0
accruedRewards = X


if X is greater than 0, then rewards will be distributed day 3, if not it will wait until accruedRewards > X in the upcoming days, and will distribute them the next day

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,270 @@
...
+
+            // get today's buyback amount
+            uint256 _accruedFee = rewardToken.balanceOf(address(this)).sub(_lastTokenBalance);
+            _lastTokenBalance = _lastTokenBalance.add(_accruedFee);
+
+            // distribute today's buyback over 1 day
+            _tokensPerSecScaled = _accruedFee.mul(PRECISION).div(1 days);
+        }
+
+        // in case `updatePool` isn't called in more than 2 days, we need to make sure user receives only one day
```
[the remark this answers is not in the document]
LouisMeMyself: Answered above

contracts/StableJoeStaking.sol
```diff
@@ -0,0 +1,270 @@
...
+        // worth of token
+        if (block.timestamp > lastRewardTimestamp + 1 days) {
+            _multiplier = 1 days;
+        } else {
+            _multiplier = block.timestamp - _lastRewardTimestamp;
+        }
```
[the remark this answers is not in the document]
LouisMeMyself: It actually is, we just don't need to update `_tokenPerSecScaled` as it's not written to memory.

Both blocks are equivalent

from `updatePool()`
```
        if (block.timestamp > lastRewardTimestamp + 1 days) {
            tokenReward = tokenReward.add(uint256(1 days).mul(tokensPerSecScaled));
            tokensPerSecScaled = 0;
        } else {
            multiplier = block.timestamp - lastRewardTimestamp;
            tokenReward = tokenReward.add(multiplier.mul(tokensPerSecScaled));
        }
```

from `pendingTokens()`
```
        if (block.timestamp > lastRewardTimestamp + 1 days) {
            _multiplier = 1 days;
        } else {
            _multiplier = block.timestamp - _lastRewardTimestamp;
        }

        _tokenReward = _tokenReward.add(_multiplier.mul(_tokensPerSecScaled));
```

test/StableJoeStaking.test.ts
```diff
@@ -0,0 +1,229 @@
...
+      await this.rewardToken.connect(this.joeMaker).transfer(this.sJoe.address, ethers.utils.parseEther("86400"))
+      expect(await this.rewardToken.balanceOf(this.sJoe.address)).to.be.equal(ethers.utils.parseEther("86400"))
+      expect(await this.sJoe.lastTokenBalance()).to.be.equal("0")
+      expect(await this.sJoe.tokenPerSec()).to.be.equal("0")
+
+      await increase(3 * 3600)
+      let block = await ethers.provider.getBlock("latest")
+
+      expect(await this.rewardToken.balanceOf(this.sJoe.address)).to.be.equal(ethers.utils.parseEther("86400"))
+      expect(await this.sJoe.pendingToken(this.alice.address)).to.be.equal(ethers.utils.parseEther("1").mul(block.timestamp % 86400))
```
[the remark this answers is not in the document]
LouisMeMyself: As we sent 86400 tokens to sJoe, `tokenPerSec = 1`
Alice is in the only one that deposited, so she owns all the rewards.

But previous day, `tokenPerSec = 0`, so she will receive no reward of day 0
and only rewards from day 1, hence `currentTimeStamp - lastDayTimestampAt0`
which is equal to `currentTimeStamp % 1 days` cause `currentTimeStamp = lastDayTimestampAt0 + elapsedSecondOfCurrentDay`

test/StableJoeStaking.test.ts
```diff
@@ -0,0 +1,229 @@
...
+      await this.sJoe.updatePool()
+      block = await ethers.provider.getBlock("latest")
+      expect(await this.sJoe.pendingToken(this.alice.address)).to.be.equal(ethers.utils.parseEther("1").mul(block.timestamp % 86400))
```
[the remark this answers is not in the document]
LouisMeMyself: I'm making sure that `pendingTokens` returns the same value even if we `updatePools`, cause that will change how `pendingTokens` will actually work, but both needs to be equal to `elapsedSecondOfCurrentDay`

LouisMeMyself:
Too complicated

PR #28, opened 2022-01-24
Setup veJOE token

0xnivek:
This PR creates a `VeJoeToken` contract following the specs in the `Staking V2 Contract Specs` notion doc.


contracts/VeERC20.sol
```diff
@@ -0,0 +1,150 @@
...
+
+import '@openzeppelin/contracts/utils/Context.sol';
+
+import './interfaces/IVeERC20.sol';
+
+/// @title VeERC20
+/// @notice Modified version of ERC20 where transfers and allowances are disabled.
+/// @dev Only minting and burning are allowed. The hook `_beforeTokenOperation` and
+/// `_afterTokenOperation` methods are called before and after minting/burning respectively.
+contract VeERC20 is Context, IVeERC20 {
```
0xnivek, line 13: note: this is a modified version of https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol

this is similar to what PTP does in https://snowtrace.io/address/0x0104eC62Afc47aF38CE214568927287E4bfDc773#code

contracts/VeERC20.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    string private _name;
+    string private _symbol;
+
+    /// @dev Emitted when `value` tokens are burned and minted
+    event Burn(address indexed account, uint256 value);
+    event Mint(address indexed beneficiary, uint256 value);
+
+    /**
+     * @dev Sets the values for {name} and {symbol}.
```
jummy123, line 26: Above you use backticks when referencing a variable and here curly brackets, is there a reason for this (maybe renders a different way?), if not would prefer to stick to one style.

contracts/VeJoeToken.sol
```diff
@@ -0,0 +1,25 @@
...
+
+pragma solidity 0.6.12;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+/// @title Vote Escrow Joe Token - veJOE
+/// @author Trader Joe
+/// @notice Infinite supply, used to receive extra farming yields and voting power
+contract VeJoeToken is ERC20("VeJoeToken", "veJOE"), Ownable {
```
jummy123, line 11: Should this not be a `VeERC20` rather than `ERC20`?

0xnivek: ah great catch! will fix

contracts/VeERC20.sol
0xnivek: Yeah this was mainly copied from [`ERC20.sol`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol) which seems to mix both

contracts/VeERC20.sol
```diff
@@ -0,0 +1,150 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.6.12;
```
jummy123, line 3: Can we use a more recent version of solidity for new contracts? This is possible with hardhat see https://hardhat.org/guides/compile-contracts.html#multiple-solidity-versions

contracts/VeERC20.sol
```diff
@@ -0,0 +1,150 @@
...
+
+    uint256 private _totalSupply;
+
+    string private _name;
+    string private _symbol;
+
+    /// @dev Emitted when `value` tokens are burned and minted
+    event Burn(address indexed account, uint256 value);
+    event Mint(address indexed beneficiary, uint256 value);
+
```
jummy123, line 24: There is no `Claim` event.

contracts/VeERC20.sol
0xnivek: ooo wow didn't know you could do this! I updated this to use `0.8.6` but one issue I ran into is that the existing `openzeppelin` contracts in the repo are using an outdated version and I can't upgrade it b/c it would break the existing contracts so I instead manually copied over the relevant latest openzeppelin contracts directly into the repo. i think this is fine but curious what you think

contracts/VeERC20.sol
```diff
@@ -0,0 +1,149 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.6;
+
+import "./openzeppelin/contracts/utils/Context.sol";
```
cryptofish7, line 5: Add this as a node_module instead

contracts/VeJoeToken.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.6;
+
+import "./openzeppelin/contracts/access/Ownable.sol";
```
cryptofish7, line 5: Use node_modules

cryptofish7 (changes requested):
Import OZ packages from node_modules instead

hardhat.config.ts
```diff
@@ -280,15 +280,6 @@ const config: HardhatUserConfig = {
...
       },
-      {
-        version: "0.7.6",
-        settings: {
-          optimizer: {
-            enabled: true,
-            runs: 200,
-          },
-        },
-      },
```
0xnivek, line 291: seems like this was duplicated from line 275-282

cryptofish7 (approved):

PR #29, opened 2022-01-24
Setup VeJoeStaking

0xnivek:
This PR sets up a basic version of `VeJoeStaking` contract following the specs in the `Staking V2 Contract Specs` notion doc which currently handles:

- Staking
- Unstaking
- Claiming
- Boosted logic (i.e. boosted generation rate, boosted duration, boosted threshold)


contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+        require(address(_veJoe) != address(0), "VeJoeStaking: unexpected zero address for _veJoe");
+        require(
+            _boostedGenerationRate > _baseGenerationRate,
+            "VeJoeStaking: expected _boostedGenerationRate to be greater than _baseGenerationRate"
+        );
+        require(_boostedThreshold <= 100, "VeJoeStaking: expected _boostedThreshold to be less than or equal to 100");
+
+        __Ownable_init();
+
+        maxCap = 100;
```
jummy123, line 81: Any reason not to set this on contract creation seeing as there is also a `setMaxCap` function.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+        baseGenerationRate = _baseGenerationRate;
+        boostedGenerationRate = _boostedGenerationRate;
+        boostedThreshold = _boostedThreshold;
+        boostedDuration = _boostedDuration;
+    }
+
+    /// @notice Set maxCap
+    /// @param _maxCap the new maxCap
+    function setMaxCap(uint256 _maxCap) external onlyOwner {
+        require(_maxCap > 0, "VeJoeStaking: expected new _maxCap to be greater than 0");
```
jummy123, line 93: What happens in the edge case where someone has more vejoe than the new `maxcap` would allow.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+        uint256 _boostedThreshold,
+        uint256 _boostedDuration
+    ) public initializer {
+        require(address(_joe) != address(0), "VeJoeStaking: unexpected zero address for _joe");
+        require(address(_veJoe) != address(0), "VeJoeStaking: unexpected zero address for _veJoe");
+        require(
+            _boostedGenerationRate > _baseGenerationRate,
+            "VeJoeStaking: expected _boostedGenerationRate to be greater than _baseGenerationRate"
+        );
+        require(_boostedThreshold <= 100, "VeJoeStaking: expected _boostedThreshold to be less than or equal to 100");
```
jummy123, line 77: `_boostedThreshold` should also be greater than 0 as this would lead to division by 0 error on line 151.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+            _claim();
+
+            userInfos[msg.sender].balance += _amount;
+
+            // User is eligible for boosted benefits if:
+            // 1. They are not already currently receiving boosted benefits
+            // 2. Their staked JOE is at least `boostedThreshold / 100 * totalStakedJoe`
+            if (userInfos[msg.sender].boostEndTimestamp == 0) {
+                uint256 totalStakedJoe = joe.balanceOf(address(this));
+                if ((userInfos[msg.sender].balance * 100) / boostedThreshold >= totalStakedJoe) {
```
jummy123, line 150: Do I understand this correctly. The minimum requirement to receive a boost it to hold at least 1% of the entire pool? This is assuming `boostedThreshold` is 1 (the minimum value).

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+                    userInfos[msg.sender].boostEndTimestamp = block.timestamp + boostedDuration;
+                }
+            }
+        } else {
+            // If the user's `lastRewardTimestamp` is 0, i.e. if this is the user's first time staking,
+            // then they will receive boosted benefits.
+            // Note that it is important we perform this check **before** we update the user's `lastRewardTimestamp`
+            // down below.
+            if (userInfos[msg.sender].lastRewardTimestamp == 0) {
+                userInfos[msg.sender].boostEndTimestamp = block.timestamp + boostedDuration;
```
jummy123, line 160: Does this leave us open to an attack where a user can switch accounts and claim a boost for a minimum amount of joe? If I am a first time user I only need to deposit a minimum amount of joe to get a boost whereas non first time users need to hold 1% of the pool.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+
+    /// @notice Get the pending amount of veJOE for a given user
+    /// @param _user The user to lookup
+    /// @return The number of pending veJOE tokens for `_user`
+    function getPendingVeJoe(address _user) public view returns (uint256) {
+        if (!getUserHasNonZeroBalance(_user)) {
+            return 0;
+        }
+
+        UserInfo storage user = userInfos[_user];
```
jummy123, line 211: You should use memory here as this is a view.

jummy123: You could also just reference directly like in some other functions and not use the `UserInfo user` variable.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+    function getUserHasNonZeroBalance(address _user) public view returns (bool) {
+        return userInfos[_user].balance > 0;
+    }
+
+    /// @dev Helper to claim any pending veJOE
+    function _claim() private {
+        uint256 veJoeToClaim = getPendingVeJoe(msg.sender);
+
+        // Update user's last reward timestamp
+        userInfos[msg.sender].lastRewardTimestamp = block.timestamp;
```
jummy123, line 260: Should this be updated if the user has 0 vejoe to claim, I think this should be moved inside the `if` statement below.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+            // If user already has staked JOE, we first send them any pending veJOE
+            _claim();
+
+            userInfos[msg.sender].balance += _amount;
+
+            // User is eligible for boosted benefits if:
+            // 1. They are not already currently receiving boosted benefits
+            // 2. Their staked JOE is at least `boostedThreshold / 100 * totalStakedJoe`
+            if (userInfos[msg.sender].boostEndTimestamp == 0) {
+                uint256 totalStakedJoe = joe.balanceOf(address(this));
```
jummy123, line 149: I think we should keep track of balance manually and increment/decrement based on deposit/withdraw functions. Now this contract can be manipulated by a user sending joe directly to the contract.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.6.12;
```
jummy123, line 3: Can we use a more recent version of solidity for new contracts? This is possible with hardhat see https://hardhat.org/guides/compile-contracts.html#multiple-solidity-versions

contracts/VeJoeStaking.sol
jummy123: We should also have a maximum value (advice is add a few 0's to the maximum you can think off) for `_maxCap`, this is to stop overflow errors.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+            _boostedGenerationRate > baseGenerationRate,
+            "VeJoeStaking: expected new _boostedGenerationRate to be greater than baseGenerationRate"
+        );
+        boostedGenerationRate = _boostedGenerationRate;
+    }
+
+    /// @notice Set boostedThreshold
+    /// @param _boostedThreshold the new boostedThreshold
+    function setBoostedThreshold(uint256 _boostedThreshold) external onlyOwner {
+        require(
```
jummy123, line 120: as per above comment we also need a > 0 check.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+
+        __Ownable_init();
+
+        maxCap = 100;
+        joe = _joe;
+        veJoe = _veJoe;
+        baseGenerationRate = _baseGenerationRate;
+        boostedGenerationRate = _boostedGenerationRate;
+        boostedThreshold = _boostedThreshold;
+        boostedDuration = _boostedDuration;
```
jummy123, line 87: Should have a maximum value validation to stop possible overflow on line 151

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    /// @notice Set baseGenerationRate
+    /// @param _baseGenerationRate the new baseGenerationRate
+    function setBaseGenerationRate(uint256 _baseGenerationRate) external onlyOwner {
+        require(
+            _baseGenerationRate < boostedGenerationRate,
+            "VeJoeStaking: expected new _baseGenerationRate to be less than boostedGenerationRate"
+        );
+        baseGenerationRate = _baseGenerationRate;
```
jummy123, line 104: It would be a "nice to have" if we had events that got emitted when important variables are updated.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+    function deposit(uint256 _amount) external {
+        require(_amount > 0, "VeJoeStaking: expected deposit amount to be greater than zero");
+
+        if (getUserHasNonZeroBalance(msg.sender)) {
+            // If user already has staked JOE, we first send them any pending veJOE
+            _claim();
+
+            userInfos[msg.sender].balance += _amount;
+
+            // User is eligible for boosted benefits if:
```
jummy123, line 145: This comment is a little confusing, I didn't know if both or only one of these conditions needs to hold.
Suggestion:
> User is eligible for boosted benefits if either of 1 or 2 are true.

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,267 @@
...
+        joe.safeTransferFrom(msg.sender, address(this), _amount);
+
+        emit Deposit(msg.sender, _amount);
+    }
+
+    /// @notice Withdraw staked JOE. Note that unstaking any amount of JOE means you will
+    /// lose all of your current veJOE.
+    /// @param _amount the amount of JOE to unstake
+    function withdraw(uint256 _amount) external {
+        require(_amount > 0, "VeJoeStaking: expected to withdraw non-zero amount of JOE");
```
jummy123, line 175: Should we be recalculating the users `boostEndTimestamp` here? Right now I could deposit a large amount of JOE to achieve a boosted generation rate and then withdraw it instantly (this flashloan).

contracts/VeJoeStaking.sol
0xnivek: yep I believe its because we can't instantiate values on upgradeable contracts

contracts/VeJoeStaking.sol
0xnivek: > What happens in the edge case where someone has more vejoe than the new maxcap would allow.

Great question! What would happen currently is they wouldn't be able to claim any more `veJoe` but if the `maxCap` were to increase in the future, they would be able to. We should probably ask murloc/mountain farmer what the desired behavior should be though.

0xnivek: > We should also have a maximum value (advice is add a few 0's to the maximum you can think off) for _maxCap, this is to stop overflow errors.

ooo fancy. I made the upper limit `100000` for now but also added a `TODO` to align with murloc/mountain farmer what the desired upper limit should be

contracts/VeJoeStaking.sol
0xnivek: I updated the logic so that `_boostedThreshold` being 0 won't cause division by 0 error. It's also probably worth considering whether we should require it be greater than 0 anyways though

contracts/VeJoeStaking.sol
0xnivek: It turns out I misunderstood this when I first wrote this two weeks ago and have updated the logic now. The correct behavior here is that the user will receive boosted benefits if they stake 5% of their current staked JOE balance and they are not already receiving boosted benefits.

For example, if the user currently has `100 JOE` staked and aren't receiving boosted benefits, they need to stake `5 JOE` to enter a boost period

contracts/VeJoeStaking.sol
0xnivek: See my comment above (https://github.com/traderjoe-xyz/joe-core/pull/29#discussion_r799954295)

contracts/VeJoeStaking.sol
0xnivek: yeah good point

contracts/VeJoeStaking.sol
0xnivek: Removed this as per https://github.com/traderjoe-xyz/joe-core/pull/29#discussion_r799954295

contracts/VeJoeStaking.sol
0xnivek: See my comment (https://github.com/traderjoe-xyz/joe-core/pull/29#discussion_r799953113)

contracts/VeJoeStaking.sol
0xnivek: agreed, added a `TODO` here to align with murloc/mountain farmer on what the upper limit should be

contracts/VeJoeStaking.sol
0xnivek: I don't believe so because anytime you withdraw `JOE`, all of your `veJOE` will be burned

contracts/VeJoeStaking.sol
0xnivek: done! see my comment in https://github.com/traderjoe-xyz/joe-core/pull/28#discussion_r799957850

contracts/VeJoeStaking.sol
jummy123: I meant pass it as an argument to `initializer`

0xnivek: ahh woops mb I misunderstood, yes we can def do that. will update

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,287 @@
...
+        }
+
+        joe.safeTransferFrom(msg.sender, address(this), _amount);
+
+        emit Deposit(msg.sender, _amount);
+    }
+
+    /// @notice Withdraw staked JOE. Note that unstaking any amount of JOE means you will
+    /// lose all of your current veJOE.
+    /// @param _amount the amount of JOE to unstake
```
cryptofish7, line 193: `the`->`The`

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,287 @@
...
+        } else {
+            // User already holds maximum amount of veJOE so there is no pending veJOE
+            return 0;
+        }
+    }
+
+    /// @notice Checks to see if a given user currently has staked JOE
+    /// @param _user the user address to check
+    /// @return whether `_user` currently has staked JOE
+    function getUserHasNonZeroBalance(address _user) private view returns (bool) {
```
cryptofish7, line 271: Rename to `_getUserHasNonZeroBalance()`

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,287 @@
...
+                return pendingVeJoe;
+            }
+        } else {
+            // User already holds maximum amount of veJOE so there is no pending veJOE
+            return 0;
+        }
+    }
+
+    /// @notice Checks to see if a given user currently has staked JOE
+    /// @param _user the user address to check
```
cryptofish7, line 269: `the`->`The`

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,287 @@
...
+    /// @return whether `_user` currently has staked JOE
```
cryptofish7, line 270: `whether`->`Whether`

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,287 @@
...
+            _claim();
+
+            uint256 userStakedJoe = userInfos[msg.sender].balance;
+
+            userInfos[msg.sender].balance += _amount;
+
+            // User is eligible for boosted benefits if and only if all of the following are true:
+            // - User is not already currently receiving boosted benefits
+            // - `_amount` is at least `boostedThreshold / 100 * userStakedJoe`
+            if (userInfos[msg.sender].boostEndTimestamp == 0 && _amount * 100 >= boostedThreshold * userStakedJoe) {
```
cryptofish7, line 171: When will `userInfos[msg.sender].boostEndTimestamp > 0` **and** user has non zero balance? Seems like this line will never be hit

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,287 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.6;
+
+import "./openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import "./openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "./openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+import "./openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
+import "./openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
```
cryptofish7, line 9: Move these into node_modules

cryptofish7 (changes requested):
Missing tests

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,287 @@
...
+contract VeJoeStaking is Initializable, OwnableUpgradeable {
+    using SafeERC20Upgradeable for IERC20Upgradeable;
+
+    struct UserInfo {
+        uint256 balance; // Amount of JOE currently staked by user
+        uint256 lastRewardTimestamp; // Timestamp of last non-zero veJOE claim, or time of first
+        // deposit if user has not claimed any veJOE yet
+        uint256 boostEndTimestamp; // Timestamp of when user stops receiving boost benefits.
+        // Note that this will be reset to 0 after the end of a boost
+    }
```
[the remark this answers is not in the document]
0xnivek: I would prefer that as well but seems like its not permitted :(
```
contracts/VeJoeStaking.sol:22:17: ParserError: Only state variables or file-level variables can have a docstring.
        uint256 balance;
                ^-----^

contracts/VeJoeStaking.sol:24:17: ParserError: Only state variables or file-level variables can have a docstring.
        uint256 lastRewardTimestamp;
                ^-----------------^

contracts/VeJoeStaking.sol:26:17: ParserError: Only state variables or file-level variables can have a docstring.
        uint256 boostEndTimestamp;
                ^---------------^

Error HH600: Compilation failed
```

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,298 @@
...
+        boostedGenerationRate = _boostedGenerationRate;
+        boostedThreshold = _boostedThreshold;
+        // TODO: Align on what the upper limit of boostedDuration should be and add require check
+        boostedDuration = _boostedDuration;
+    }
+
+    /// @notice Set maxCap
+    /// @param _maxCap The new maxCap
+    function setMaxCap(uint256 _maxCap) external onlyOwner {
+        // TODO: Align on what the upper limit of maxCap should be
```
jummy123, line 106: I think we should add a validation here `require(_maxCap > maxCap)` this is to ensure no user has more veJOE than their maxCap would allow. If this were the case future changes to economic parameters could be thrown out by this. I have also asked MrBeaverTail and he agrees this number should never decrease.

cryptofish7 (changes requested):
Waiting for `lastRewardTimestamp < boostEndTimestamp < now` case

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,340 @@
...
+    function deposit(uint256 _amount) external {
+        require(_amount > 0, "VeJoeStaking: expected deposit amount to be greater than zero");
+
+        UserInfo storage userInfo = userInfos[msg.sender];
+
+        if (_getUserHasNonZeroBalance(msg.sender)) {
+            // If user already has staked JOE, we first send them any pending veJOE
+            _claim();
+
+            uint256 userStakedJoe = userInfos[msg.sender].balance;
```
cryptofish7, line 174: Change to `uint256 userStakedJoe = userInfo.balance;`

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,340 @@
...
+
+            userInfo.balance = userInfo.balance.add(_amount);
```
cryptofish7, line 176: Change to `userInfo.balance = userStakedJoe.add(_amount);`

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,340 @@
...
+                //    `user.boostEndTimestamp < user.lastRewardTimestamp`.
+                // 2. That means that at time `user.lastRewardTimestamp`, the user claimed
+                //    some veJOE. Furthermore, we know that anytime a user claims some veJOE,
+                //    if the current timestamp is greater than or equal to `user.boostEndTimestamp`,
+                //    we will update `user.boostEndTimestamp` to be `0` (see `_claim` method).
+                // 3. This means that `user.boostEndTimestamp` should be `0` but that contradicts our
+                //    assumption that `user.boostEndTimestamp != 0`
+                // QED.
+                // With this, we now know `0 < user.lastRewardTimestamp <= user.boostEndTimestamp < block.timestamp`,
+                // which will allow us to perform the following logic safely.
```
cryptofish7, line 276: Nice

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,340 @@
...
+                uint256 boostedAccVeJoePerJoe = boostedTimeElapsed.mul(boostedGenerationRate).div(PRECISION);
+                uint256 boostedPendingVeJoe = boostedAccVeJoePerJoe.mul(user.balance);
+
+                uint256 baseTimeElapsed = block.timestamp.sub(user.boostEndTimestamp);
+                uint256 baseAccVeJoePerVeJoe = baseTimeElapsed.mul(baseGenerationRate).div(PRECISION);
+                uint256 basePendingVeJoe = baseAccVeJoePerVeJoe.mul(user.balance);
+
+                pendingVeJoe = boostedPendingVeJoe.add(basePendingVeJoe);
+            } else {
+                uint256 accVeJoePerJoe = secondsElapsed.mul(baseGenerationRate).div(PRECISION);
```
cryptofish7, line 293: I'd make a comment here that this is the case where boost has ended

cryptofish7 (changes requested):
I'm pretty sure `PRECISION` is not used in the correct way here and code will be liable to rounding errors (@LouisMeMyself can you please review this part pls too).

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,340 @@
...
+                // those `boostEndTimestamp - lastRewardTimestamp` seconds.
+                uint256 boostedTimeElapsed = user.boostEndTimestamp.sub(user.lastRewardTimestamp);
+                uint256 boostedAccVeJoePerJoe = boostedTimeElapsed.mul(boostedGenerationRate).div(PRECISION);
+                uint256 boostedPendingVeJoe = boostedAccVeJoePerJoe.mul(user.balance);
+
+                uint256 baseTimeElapsed = block.timestamp.sub(user.boostEndTimestamp);
+                uint256 baseAccVeJoePerVeJoe = baseTimeElapsed.mul(baseGenerationRate).div(PRECISION);
+                uint256 basePendingVeJoe = baseAccVeJoePerVeJoe.mul(user.balance);
+
+                pendingVeJoe = boostedPendingVeJoe.add(basePendingVeJoe);
```
LouisMeMyself, line 291: pendingVeJoe = boostedPendingVeJoe.add(basePendingVeJoe).div(PRECISION)

Now we divide it because those 2 numbers are now in 1e36 (18 + `veJOE.decimals()`), and we need them to have `veJOE.decimals()` decimals.

It may be easier to understand if we divide at L289 and L285, but it would cost twice more gas than doing it here

contracts/VeJoeStaking.sol
LouisMeMyself, line 293: `uint256 accVeJoePerJoe = secondsElapsed.mul(baseGenerationRate)`
same reason

contracts/VeJoeStaking.sol
```diff
@@ -0,0 +1,340 @@
...
+            } else {
+                uint256 accVeJoePerJoe = secondsElapsed.mul(baseGenerationRate).div(PRECISION);
+                pendingVeJoe = accVeJoePerJoe.mul(user.balance);
```
LouisMeMyself, line 294: `pendingVeJoe = accVeJoePerJoe.mul(user.balance).div(PRECISION)`

LouisMeMyself (changes requested):

cryptofish7 (approved):
LGTM

PR #36, opened 2022-02-09
Use boosting to share rewards

jummy123:
I started this code by copying `MasterChefJoeV3` and modifying the necessary functions for boosted staking. I know it is not clear from the diff but a lot of the code is copied verbatim from MCJV3. This means we have a couple of choices to make with tradeoffs on both sides.

* Changing the solidity version
* Refactoring/renaming unmodified code from MCJV3
* Rewrite/cleanup all the copied natspec (puncutation etc) 

I'm erring on the side of renaming variables and error messages to our standard conventions but not modifying the solidity version or unchanged functions. Open for discussion :face_with_spiral_eyes:

### DIFF
--- a/contracts/BoostedMasterChefJoe.sol
+++ b/contracts/BoostedMasterChefJoe.sol
@@ -199,7 +199,7 @@ contract BoostedMasterChefJoe is Initializable, OwnableUpgradeable, ReentrancyGu
         bool _overwrite
     ) external onlyOwner {
         PoolInfo memory pool = poolInfo[_pid];
-        totalAllocPoint = totalAllocPoint.sub(poolInfo[_pid].allocPoint).add(_allocPoint);
+        totalAllocPoint = totalAllocPoint.sub(pool.allocPoint).add(_allocPoint);
         pool.allocPoint = _allocPoint;
         if (_overwrite) {
             if (address(_rewarder) != address(0)) {
--- a/contracts/BoostedMasterChefJoe.sol
+++ b/contracts/BoostedMasterChefJoe.sol
@@ -152,7 +152,6 @@ contract BoostedMasterChefJoe is Initializable, OwnableUpgradeable, ReentrancyGu
     }
 
     /// @notice Add a new LP to the pool. Can only be called by the owner.
-    /// DO NOT add the same LP token more than once. Rewards will be messed up if you do.
     /// @param _allocPoint AP of the new pool.
     /// @param _lpToken Address of the LP ERC-20 token.
     /// @param _rewarder Address of the rewarder delegate.
--- a/contracts/BoostedMasterChefJoe.sol
+++ b/contracts/BoostedMasterChefJoe.sol
@@ -101,7 +101,7 @@ contract BoostedMasterChefJoe is Ownable, ReentrancyGuard {
     event UpdatePool(uint256 indexed pid, uint256 lastRewardTimestamp, uint256 lpSupply, uint256 accJoePerShare);
     event Harvest(address indexed user, uint256 indexed pid, uint256 amount);
     event EmergencyWithdraw(address indexed user, uint256 indexed pid, uint256 amount);
-    event Init();
+    event Init(uint256 amount);
     event MaxBoostFactorChanged(uint256 maxBoostFactor);
 
     /// @param _MASTER_CHEF_V2 The MCJV2 contract address
@@ -135,7 +135,7 @@ contract BoostedMasterChefJoe is Ownable, ReentrancyGuard {
         _dummyToken.safeTransferFrom(msg.sender, address(this), balance);
         _dummyToken.approve(address(MASTER_CHEF_V2), balance);
         MASTER_CHEF_V2.deposit(MASTER_PID, balance);
-        emit Init();
+        emit Init(balance);
     }
 
     /// @notice Returns the number of BMCJ pools.
--- a/contracts/BoostedMasterChefJoe.sol
+++ b/contracts/BoostedMasterChefJoe.sol
@@ -9,6 +9,7 @@ import "@openzeppelin/contracts/utils/EnumerableSet.sol";
 import "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
 import "./interfaces/IERC20.sol";
 import "./interfaces/IRewarder.sol";
+import "./libraries/BoringJoeERC20.sol";
 
 interface IMasterChef {
     struct UserInfo {
@@ -38,89 +39,6 @@ interface IMasterChef {
     function investorPercent() external view returns (uint256);
 }
 
-
-library BoringERC20 {
-    bytes4 private constant SIG_SYMBOL = 0x95d89b41; // symbol()
-    bytes4 private constant SIG_NAME = 0x06fdde03; // name()
-    bytes4 private constant SIG_DECIMALS = 0x313ce567; // decimals()
-    bytes4 private constant SIG_TRANSFER = 0xa9059cbb; // transfer(address,uint256)
-    bytes4 private constant SIG_TRANSFER_FROM = 0x23b872dd; // transferFrom(address,address,uint256)
-
-    function returnDataToString(bytes memory data) internal pure returns (string memory) {
-        if (data.length >= 64) {
-            return abi.decode(data, (string));
-        } else if (data.length == 32) {
-            uint8 i = 0;
-            while (i < 32 && data[i] != 0) {
-                i++;
-            }
-            bytes memory bytesArray = new bytes(i);
-            for (i = 0; i < 32 && data[i] != 0; i++) {
-                bytesArray[i] = data[i];
-            }
-            return string(bytesArray);
-        } else {
-            return "???";
-        }
-    }
-
-    /// @notice Provides a safe ERC20.symbol version which returns '???' as fallback string.
-    /// @param token The address of the ERC-20 token contract.
-    /// @return (string) Token symbol.
-    function safeSymbol(IERC20 token) internal view returns (string memory) {
-        (bool success, bytes memory data) = address(token).staticcall(abi.encodeWithSelector(SIG_SYMBOL));
-        return success ? returnDataToString(data) : "???";
-    }
-
-    /// @notice Provides a safe ERC20.name version which returns '???' as fallback string.
-    /// @param token The address of the ERC-20 token contract.
-    /// @return (string) Token name.
-    function safeName(IERC20 token) internal view returns (string memory) {
-        (bool success, bytes memory data) = address(token).staticcall(abi.encodeWithSelector(SIG_NAME));
-        return success ? returnDataToString(data) : "???";
-    }
-
-    /// @notice Provides a safe ERC20.decimals version which returns '18' as fallback value.
-    /// @param token The address of the ERC-20 token contract.
-    /// @return (uint8) Token decimals.
-    function safeDecimals(IERC20 token) internal view returns (uint8) {
-        (bool success, bytes memory data) = address(token).staticcall(abi.encodeWithSelector(SIG_DECIMALS));
-        return success && data.length == 32 ? abi.decode(data, (uint8)) : 18;
-    }
-
-    /// @notice Provides a safe ERC20.transfer version for different ERC-20 implementations.
-    /// Reverts on a failed transfer.
-    /// @param token The address of the ERC-20 token.
-    /// @param to Transfer tokens to.
-    /// @param amount The token amount.
-    function safeTransfer(
-        IERC20 token,
-        address to,
-        uint256 amount
-    ) internal {
-        (bool success, bytes memory data) = address(token).call(abi.encodeWithSelector(SIG_TRANSFER, to, amount));
-        require(success && (data.length == 0 || abi.decode(data, (bool))), "BoringERC20: Transfer failed");
-    }
-
-    /// @notice Provides a safe ERC20.transferFrom version for different ERC-20 implementations.
-    /// Reverts on a failed transfer.
-    /// @param token The address of the ERC-20 token.
-    /// @param from Transfer tokens from.
-    /// @param to Transfer tokens to.
-    /// @param amount The token amount.
-    function safeTransferFrom(
-        IERC20 token,
-        address from,
-        address to,
-        uint256 amount
-    ) internal {
-        (bool success, bytes memory data) = address(token).call(
-            abi.encodeWithSelector(SIG_TRANSFER_FROM, from, to, amount)
-        );
-        require(success && (data.length == 0 || abi.decode(data, (bool))), "BoringERC20: TransferFrom failed");
-    }
-}
-
 /// @notice The (older) MasterChefJoeV2 contract gives out a constant number of JOE
 /// tokens per block.  It is the only address with minting rights for JOE.  The idea
 /// for this BoostedMasterChefJoe (BMCJ) contract is therefore to be the owner of a
@@ -139,7 +57,7 @@ library BoringERC20 {
 /// participants. Total liquidity is the sum of all boosted liquidity.
 contract BoostedMasterChefJoe is Ownable, ReentrancyGuard {
     using SafeMath for uint256;
-    using BoringERC20 for IERC20;
+    using BoringJoeERC20 for IERC20;
     using EnumerableSet for EnumerableSet.AddressSet;
 
     /// @notice Info of each BMCJ user
--- a/contracts/BoostedMasterChefJoe.sol
+++ b/contracts/BoostedMasterChefJoe.sol
@@ -106,7 +106,6 @@ contract BoostedMasterChefJoe is Initializable, OwnableUpgradeable, ReentrancyGu
     event Harvest(address indexed user, uint256 indexed pid, uint256 amount);
     event EmergencyWithdraw(address indexed user, uint256 indexed pid, uint256 amount);
     event Init(uint256 amount);
-    event MaxBoostFactorChanged(uint256 maxBoostFactor);
 
     /// @param _MASTER_CHEF_V2 The MCJV2 contract address
     /// @param _joe The JOE token contract address
--- a/contracts/BoostedMasterChefJoe.sol
+++ b/contracts/BoostedMasterChefJoe.sol
@@ -1,0 +1,589 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.6.12;
+pragma experimental ABIEncoderV2;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/math/SafeMath.sol";
+import "@openzeppelin/contracts/utils/EnumerableSet.sol";
+import "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
+import "./interfaces/IERC20.sol";
+
+interface IMasterChef {
+    struct UserInfo {
+        uint256 amount; // How many LP tokens the user has provided.
+        uint256 rewardDebt; // Reward debt. See explanation below.
+    }
+
+    struct PoolInfo {
+        IERC20 lpToken; // Address of LP token contract.
+        uint256 allocPoint; // How many allocation points assigned to this pool. JOE to distribute per block.
+        uint256 lastRewardTimestamp; // Last block number that JOE distribution occurs.
+        uint256 accJoePerShare; // Accumulated JOE per share, times 1e12. See below.
+    }
+
+    function poolInfo(uint256 pid) external view returns (IMasterChef.PoolInfo memory);
+
+    function totalAllocPoint() external view returns (uint256);
+
+    function joePerSec() external view returns (uint256);
+
+    function deposit(uint256 _pid, uint256 _amount) external;
+
+    function devPercent() external view returns (uint256);
+
+    function treasuryPercent() external view returns (uint256);
+
+    function investorPercent() external view returns (uint256);
+}
+
+interface IRewarder {
+    function onJoeReward(address user, uint256 newLpAmount) external;
+
+    function pendingTokens(address user) external view returns (uint256 pending);
+
+    function rewardToken() external view returns (IERC20);
+}
+
+library BoringERC20 {
+    bytes4 private constant SIG_SYMBOL = 0x95d89b41; // symbol()
+    bytes4 private constant SIG_NAME = 0x06fdde03; // name()
+    bytes4 private constant SIG_DECIMALS = 0x313ce567; // decimals()
+    bytes4 private constant SIG_TRANSFER = 0xa9059cbb; // transfer(address,uint256)
+    bytes4 private constant SIG_TRANSFER_FROM = 0x23b872dd; // transferFrom(address,address,uint256)
+
+    function returnDataToString(bytes memory data) internal pure returns (string memory) {
+        if (data.length >= 64) {
+            return abi.decode(data, (string));
+        } else if (data.length == 32) {
+            uint8 i = 0;
+            while (i < 32 && data[i] != 0) {
+                i++;
+            }
+            bytes memory bytesArray = new bytes(i);
+            for (i = 0; i < 32 && data[i] != 0; i++) {
+                bytesArray[i] = data[i];
+            }
+            return string(bytesArray);
+        } else {
+            return "???";
+        }
+    }
+
+    /// @notice Provides a safe ERC20.symbol version which returns '???' as fallback string.
+    /// @param token The address of the ERC-20 token contract.
+    /// @return (string) Token symbol.
+    function safeSymbol(IERC20 token) internal view returns (string memory) {
+        (bool success, bytes memory data) = address(token).staticcall(abi.encodeWithSelector(SIG_SYMBOL));
+        return success ? returnDataToString(data) : "???";
+    }
+
+    /// @notice Provides a safe ERC20.name version which returns '???' as fallback string.
+    /// @param token The address of the ERC-20 token contract.
+    /// @return (string) Token name.
+    function safeName(IERC20 token) internal view returns (string memory) {
+        (bool success, bytes memory data) = address(token).staticcall(abi.encodeWithSelector(SIG_NAME));
+        return success ? returnDataToString(data) : "???";
+    }
+
+    /// @notice Provides a safe ERC20.decimals version which returns '18' as fallback value.
+    /// @param token The address of the ERC-20 token contract.
+    /// @return (uint8) Token decimals.
+    function safeDecimals(IERC20 token) internal view returns (uint8) {
+        (bool success, bytes memory data) = address(token).staticcall(abi.encodeWithSelector(SIG_DECIMALS));
+        return success && data.length == 32 ? abi.decode(data, (uint8)) : 18;
+    }
+
+    /// @notice Provides a safe ERC20.transfer version for different ERC-20 implementations.
+    /// Reverts on a failed transfer.
+    /// @param token The address of the ERC-20 token.
+    /// @param to Transfer tokens to.
+    /// @param amount The token amount.
+    function safeTransfer(
+        IERC20 token,
+        address to,
+        uint256 amount
+    ) internal {
+        (bool success, bytes memory data) = address(token).call(abi.encodeWithSelector(SIG_TRANSFER, to, amount));
+        require(success && (data.length == 0 || abi.decode(data, (bool))), "BoringERC20: Transfer failed");
+    }
+
+    /// @notice Provides a safe ERC20.transferFrom version for different ERC-20 implementations.
+    /// Reverts on a failed transfer.
+    /// @param token The address of the ERC-20 token.
+    /// @param from Transfer tokens from.
+    /// @param to Transfer tokens to.
+    /// @param amount The token amount.
+    function safeTransferFrom(
+        IERC20 token,
+        address from,
+        address to,
+        uint256 amount
+    ) internal {
+        (bool success, bytes memory data) = address(token).call(
+            abi.encodeWithSelector(SIG_TRANSFER_FROM, from, to, amount)
+        );
+        require(success && (data.length == 0 || abi.decode(data, (bool))), "BoringERC20: TransferFrom failed");
+    }
+}
+
+/// @notice The (older) MasterChefJoeV2 contract gives out a constant number of JOE
+/// tokens per block.  It is the only address with minting rights for JOE.  The idea
+/// for this BoostedMasterChefJoe (BMCJ) contract is therefore to be the owner of a
+/// dummy token that is deposited into the MasterChefJoeV2 (MCJV2) contract.  The
+/// allocation point for this pool on MCJV2 is the total allocation point for all
+/// pools on BMCJ.
+///
+/// This MasterChef also skews how many rewards users receive, it does this by
+/// modifying the algorithm that calculates how many tokens are rewarded to
+/// depositors. Whereas MasterChef calculates rewards based on emission rate and
+/// total liquidity, this version uses adjusted parameters to this calculation.
+///
+/// A users `boostedAmount` (liquidity multiplier) is calculated by the actual supplied
+/// liquidity multiplied by a boost factor. The boost factor is calculated by the
+/// amount of veJOE held by the user over the total veJOE amount held by all pool
+/// participants. Total liquidity is the sum of all boosted liquidity.
+contract BoostedMasterChefJoe is Ownable, ReentrancyGuard {
+    using SafeMath for uint256;
+    using BoringERC20 for IERC20;
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    /// @notice Info of each BMCJ user
+    /// `amount` LP token amount the user has provided
+    /// `rewardDebt` The amount of JOE entitled to the user
+    /// `veJoeBalance` the users balance of veJOE token. This needs
+    /// to be stored so we have a prior value when updating the `PoolInfo`
+    struct UserInfo {
+        uint256 amount;
+        uint256 rewardDebt;
+        uint256 veJoeBalance;
+    }
+
+    /// @notice Info of each BMCJ pool
+    /// `allocPoint` The amount of allocation points assigned to the pool
+    /// Also known as the amount of JOE to distribute per block
+    struct PoolInfo {
+        IERC20 lpToken;
+        uint256 accJoePerShare;
+        uint256 lastRewardTimestamp;
+        uint256 allocPoint;
+        IRewarder rewarder;
+        // The sum of all veJoe held by users participating in this farm
+        // This value is updated when
+        // - A user enter/leaves a farm
+        // - A user claims veJOE
+        // - A user unstakes JOE
+        uint256 totalVeJoe;
+        // The total boosted `amount` on the farm
+        // This is the sum of all users boosted amounts in the farm. Updated when
+        // totalVeJoe is updated
+        // This is used instead of the usual `lpToken.balanceOf(address(this))`
+        uint256 totalBoostedAmount;
+        // Even though it seems redundant to keep track of two variables here
+        // the `totalBoostedAmount` means we don't need to iterate over all
+        // depositors to calculate the pending reward.
+    }
+
+    /// @notice Address of MCJV2 contract
+    IMasterChef public immutable MASTER_CHEF_V2;
+    /// @notice Address of JOE contract
+    IERC20 public immutable JOE;
+    /// @notice Address of veJOE contract
+    IERC20 public immutable VEJOE;
+    /// @notice The index of BMCJ master pool in MCJV2
+    uint256 public immutable MASTER_PID;
+    /// @notice Info of each BMCJ pool
+    PoolInfo[] public poolInfo;
+    // Set of all LP tokens that have been added as pools
+    EnumerableSet.AddressSet private lpTokens;
+    /// @notice Info of each user that stakes LP tokens
+    mapping(uint256 => mapping(address => UserInfo)) public userInfo;
+    /// @dev Total allocation points. Must be the sum of all allocation points in all pools
+    uint256 public totalAllocPoint;
+    uint256 private constant ACC_TOKEN_PRECISION = 1e18;
+    /// @dev A maximum scaling factor applied to boosted amounts
+    uint256 public maxBoostFactor = 1500;
+    /// @dev Amount of claimable Joe the user has, this is required as we
+    /// need to update rewardDebt after a token operation but we don't
+    /// want to send a reward at this point. This amount gets added onto
+    /// the pending amount when a user claims
+    mapping(uint256 => mapping(address => uint256)) public claimableJoe;
+
+    event Add(uint256 indexed pid, uint256 allocPoint, IERC20 indexed lpToken, IRewarder indexed rewarder);
+    event Set(uint256 indexed pid, uint256 allocPoint, IRewarder indexed rewarder, bool overwrite);
+    event Deposit(address indexed user, uint256 indexed pid, uint256 amount);
+    event Withdraw(address indexed user, uint256 indexed pid, uint256 amount);
+    event UpdatePool(uint256 indexed pid, uint256 lastRewardTimestamp, uint256 lpSupply, uint256 accJoePerShare);
+    event Harvest(address indexed user, uint256 indexed pid, uint256 amount);
+    event EmergencyWithdraw(address indexed user, uint256 indexed pid, uint256 amount);
+    event Init();
+    event MaxBoostFactorChanged(uint256 maxBoostFactor);
+
+    /// @param _MASTER_CHEF_V2 The MCJV2 contract address
+    /// @param _joe The JOE token contract address
+    /// @param _veJoe The veJOE token contract address
+    /// @param _MASTER_PID The pool ID of the dummy token on the base MCJV2 contract
+    constructor(
+        IMasterChef _MASTER_CHEF_V2,
+        IERC20 _joe,
+        IERC20 _veJoe,
+        uint256 _MASTER_PID
+    ) public {
+        MASTER_CHEF_V2 = _MASTER_CHEF_V2;
+        JOE = _joe;
+        VEJOE = _veJoe;
+        MASTER_PID = _MASTER_PID;
+    }
+
+    /// @notice Deposits a dummy token to `MASTER_CHEF_V2` MCJV2. This is required because MCJV2
+    /// holds the minting rights for JOE.  Any balance of transaction sender in `_dummyToken` is transferred.
+    /// The allocation point for the pool on MCJV2 is the total allocation point for all pools that receive
+    /// double incentives.
+    /// @param _dummyToken The address of the ERC-20 token to deposit into MCJV2.
+    function init(IERC20 _dummyToken) external onlyOwner {
+        uint256 balance = _dummyToken.balanceOf(msg.sender);
+        require(balance != 0, "BoostedMasterChefJoe: Balance must exceed 0");
+        _dummyToken.safeTransferFrom(msg.sender, address(this), balance);
+        _dummyToken.approve(address(MASTER_CHEF_V2), balance);
+        MASTER_CHEF_V2.deposit(MASTER_PID, balance);
+        emit Init();
+    }
+
+    /// @notice Returns the number of BMCJ pools.
+    /// @return pools The amount of pools in this farm
+    function poolLength() external view returns (uint256 pools) {
+        pools = poolInfo.length;
+    }
+
+    /// @notice Add a new LP to the pool. Can only be called by the owner.
+    /// DO NOT add the same LP token more than once. Rewards will be messed up if you do.
+    /// @param _allocPoint AP of the new pool.
+    /// @param _lpToken Address of the LP ERC-20 token.
+    /// @param _rewarder Address of the rewarder delegate.
+    function add(
+        uint256 _allocPoint,
+        IERC20 _lpToken,
+        IRewarder _rewarder
+    ) external onlyOwner {
+        require(!lpTokens.contains(address(_lpToken)), "BoostedMasterChefJoe: LP already added");
+        require(poolInfo.length <= 50, "BoostedMasterChefJoe: Too many pools");
+        // Sanity check to ensure _lpToken is an ERC20 token
+        _lpToken.balanceOf(address(this));
+        // Sanity check if we add a rewarder
+        if (address(_rewarder) != address(0)) {
+            _rewarder.onJoeReward(address(0), 0);
+        }
+
+        uint256 lastRewardTimestamp = block.timestamp;
+        totalAllocPoint = totalAllocPoint.add(_allocPoint);
+
+        poolInfo.push(
+            PoolInfo({
+                lpToken: _lpToken,
+                allocPoint: _allocPoint,
+                lastRewardTimestamp: lastRewardTimestamp,
+                accJoePerShare: 0,
+                rewarder: _rewarder,
+                totalVeJoe: 0,
+                totalBoostedAmount: 0
+            })
+        );
+        lpTokens.add(address(_lpToken));
+        emit Add(poolInfo.length.sub(1), _allocPoint, _lpToken, _rewarder);
+    }
+
+    /// @notice Update the given pool's JOE allocation point and `IRewarder` contract. Can only be called by the owner.
+    /// @param _pid The index of the pool. See `poolInfo`
+    /// @param _allocPoint New AP of the pool
+    /// @param _rewarder Address of the rewarder delegate
+    /// @param _overwrite True if _rewarder should be `set`. Otherwise `_rewarder` is ignored
+    function set(
+        uint256 _pid,
+        uint256 _allocPoint,
+        IRewarder _rewarder,
+        bool _overwrite
+    ) external onlyOwner {
+        PoolInfo memory pool = poolInfo[_pid];
+        totalAllocPoint = totalAllocPoint.sub(poolInfo[_pid].allocPoint).add(_allocPoint);
+        pool.allocPoint = _allocPoint;
+        if (_overwrite) {
+            _rewarder.onJoeReward(address(0), 0); // sanity check
+            pool.rewarder = _rewarder;
+        }
+        poolInfo[_pid] = pool;
+        emit Set(_pid, _allocPoint, _overwrite ? _rewarder : pool.rewarder, _overwrite);
+    }
+
+    /// @notice View function to see pending JOE on frontend
+    /// @param _pid The index of the pool. See `poolInfo`
+    /// @param _user Address of user
+    /// @return pendingJoe JOE reward for a given user.
+    /// @return bonusTokenAddress The address of the bonus reward.
+    /// @return bonusTokenSymbol The symbol of the bonus token.
+    /// @return pendingBonusToken The amount of bonus rewards pending.
+    function pendingTokens(uint256 _pid, address _user)
+        external
+        view
+        returns (
+            uint256 pendingJoe,
+            address bonusTokenAddress,
+            string memory bonusTokenSymbol,
+            uint256 pendingBonusToken
+        )
+    {
+        PoolInfo memory pool = poolInfo[_pid];
+        UserInfo storage user = userInfo[_pid][_user];
+        uint256 accJoePerShare = pool.accJoePerShare;
+
+        if (block.timestamp > pool.lastRewardTimestamp && pool.totalBoostedAmount != 0) {
+            uint256 secondsElapsed = block.timestamp.sub(pool.lastRewardTimestamp);
+            uint256 joeReward = secondsElapsed.mul(joePerSec()).mul(pool.allocPoint).div(totalAllocPoint);
+            accJoePerShare = accJoePerShare.add(joeReward.mul(ACC_TOKEN_PRECISION).div(pool.totalBoostedAmount));
+        }
+
+        uint256 userLiquidity = getBoostedLiquidity(_pid, _user);
+        pendingJoe = userLiquidity.mul(accJoePerShare).div(ACC_TOKEN_PRECISION).sub(user.rewardDebt).add(
+            claimableJoe[_pid][_user]
+        );
+
+        // If it's a double reward farm, we return info about the bonus token
+        if (address(pool.rewarder) != address(0)) {
+            bonusTokenAddress = address(pool.rewarder.rewardToken());
+            bonusTokenSymbol = IERC20(pool.rewarder.rewardToken()).safeSymbol();
+            pendingBonusToken = pool.rewarder.pendingTokens(_user);
+        }
+    }
+
+    /// @notice Update reward variables for all pools. Be careful of gas spending!
+    /// @param _pids Pool IDs of all to be updated. Make sure to update all active pools
+    function massUpdatePools(uint256[] calldata _pids) external {
+        uint256 len = _pids.length;
+        for (uint256 i = 0; i < len; ++i) {
+            updatePool(_pids[i]);
+        }
+    }
+
+    /// @notice Calculates and returns the `amount` of JOE per second
+    /// @return _amount The amount of JOE emitted per second
+    function joePerSec() public view returns (uint256 _amount) {
+        uint256 total = 1000;
+        uint256 lpPercent = total.sub(MASTER_CHEF_V2.devPercent()).sub(MASTER_CHEF_V2.treasuryPercent()).sub(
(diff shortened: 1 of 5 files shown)


contracts/BoostedMasterChef.sol
```diff
@@ -404,6 +524,70 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
+            );
+        }
+    }
+
+    /// @notice Returns a users boosted emission share
+    /// @param _pid The index of the pool. See `poolInfo`
+    /// @param _user The address of the depositor
+    function getUserBoost(uint256 _pid, address _user) public view returns (uint256 userBoost) {
+        PoolInfo memory pool = poolInfo[_pid];
+        UserInfo storage user = userInfo[_pid][_user];
```
cryptofish7, line 584: Use memory

contracts/BoostedMasterChef.sol
```diff
@@ -288,13 +338,23 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
-            accJoePerShare = accJoePerShare.add(joeReward.mul(ACC_TOKEN_PRECISION).div(lpSupply));
+            accJoePerShare = accJoePerShare.add(joeReward.mul(ACC_TOKEN_PRECISION).div(pool.totalBoostedAmount));
         }
-        pendingJoe = user.amount.mul(accJoePerShare).div(ACC_TOKEN_PRECISION).sub(user.rewardDebt);
+
+        // TODO: Precision.
+        uint256 userBoost = 1;
+        uint256 userVeJoeBalance = VEJOE.balanceOf(_user);
+        if (userVeJoeBalance > 0) {
+            userBoost = userBoost.add(userVeJoeBalance.div(pool.totalVeJoe)).mul(maxBoostFactor);
```
cryptofish7, line 352: Change to userBoost + (userVeJoeBalance / totalVeJoe) * maxBoostFactor

cryptofish7: Also, why not use `getUserBoost()` here?

contracts/BoostedMasterChef.sol
```diff
@@ -381,17 +470,48 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
         // Effects
+
+        // If the user has left the pool altogether we subtract all their
+        // veJoe balance from `pool.totalVeJoe`, and full boosted balance
+        // from `pool.totalBoostedAmount` otherwise we just adjust down the
+        // total boosted amount as total vejoe is unchanged.
+
+        // Update the `pool.totalBoostedAmount`.
+        if (user.amount.sub(amount) == 0) {
+            pool.totalVeJoe = pool.totalVeJoe.sub(VEJOE.balanceOf(msg.sender));
```
cryptofish7, line 499: Why not subtract with `userInfo.veJoeBalance` here? What if user increases his veJOE balance since it was updated in BCMJ?

contracts/BoostedMasterChef.sol
```diff
@@ -381,17 +470,48 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
+            pool.totalBoostedAmount = pool.totalBoostedAmount.sub(user.amount * userBoost);
+            // Is this needed?
+            user.veJoeBalance = 0;
+        } else {
+            pool.totalBoostedAmount = pool.totalBoostedAmount.sub(user.amount.mul(userBoost)).add(
+                user.amount.sub(amount).mul(userBoost)
+            );
+            pool.totalVeJoe = pool.totalVeJoe.sub(user.veJoeBalance).add(VEJOE.balanceOf(msg.sender));
```
cryptofish7, line 507: Is it possible to get negative value here for `pool.totalVeJoe`?

contracts/BoostedMasterChef.sol
```diff
@@ -381,17 +470,48 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
+            user.veJoeBalance = VEJOE.balanceOf(msg.sender);
+        }
+
         user.amount = user.amount.sub(amount);
-        user.rewardDebt = user.amount.mul(pool.accJoePerShare).div(ACC_TOKEN_PRECISION);
+        user.rewardDebt = user.amount.mul(getUserBoost(pid, msg.sender)).mul(pool.accJoePerShare).div(
```
cryptofish7, line 512: Use `userBoost` from L473

contracts/BoostedMasterChef.sol
```diff
@@ -404,6 +524,70 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
+                .sub(user.rewardDebt);
+
+            // Increase claimableJoe
+            claimableJoe[pid][_user] += pending;
+
+            // Update pool.totalVeJoe
+            pool.totalVeJoe = pool.totalVeJoe.sub(user.veJoeBalance).add(_newUserBalance);
+
+            // Decrement the total boosted amount and increment it after updating the
+            // users veJoeBalance .
```
cryptofish7, line 564: Remove trailing `.`

contracts/BoostedMasterChef.sol
```diff
@@ -404,6 +524,70 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
+            pool.totalBoostedAmount = pool.totalBoostedAmount.sub(user.amount.mul(getUserBoost(pid, _user)));
+
+            // Update users veJoeBalance.
+            user.veJoeBalance = _newUserBalance;
+
+            // Update the pools total boosted amount.
+            pool.totalBoostedAmount = pool.totalBoostedAmount.add(user.amount.mul(getUserBoost(pid, _user)));
+            // update reward debt, take into account new boost
```
cryptofish7, line 572: `update`->`Update`

contracts/BoostedMasterChef.sol
```diff
@@ -404,6 +524,70 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
+            claimableJoe[pid][_user] += pending;
+
+            // Update pool.totalVeJoe
+            pool.totalVeJoe = pool.totalVeJoe.sub(user.veJoeBalance).add(_newUserBalance);
+
+            // Decrement the total boosted amount and increment it after updating the
+            // users veJoeBalance .
+            pool.totalBoostedAmount = pool.totalBoostedAmount.sub(user.amount.mul(getUserBoost(pid, _user)));
+
+            // Update users veJoeBalance.
```
cryptofish7, line 567: Remove trailing `.`

contracts/BoostedMasterChef.sol
```diff
@@ -404,6 +524,70 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
+            user.veJoeBalance = _newUserBalance;
+
+            // Update the pools total boosted amount.
```
cryptofish7, line 570: Remove trailing `.`

contracts/BoostedMasterChef.sol
```diff
@@ -404,6 +524,70 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
+            pool.totalBoostedAmount = pool.totalBoostedAmount.add(user.amount.mul(getUserBoost(pid, _user)));
+            // update reward debt, take into account new boost
+            user.rewardDebt = user.amount.mul(getUserBoost(pid, _user)).mul(pool.accJoePerShare).div(
```
cryptofish7, line 573: Can cache `getUserBoost()` here with the call to L571

contracts/BoostedMasterChef.sol
```diff
@@ -404,6 +524,70 @@ contract MasterChefJoeV3 is Ownable, ReentrancyGuard {
...
+    function getUserBoost(uint256 _pid, address _user) public view returns (uint256 userBoost) {
+        PoolInfo memory pool = poolInfo[_pid];
+        UserInfo storage user = userInfo[_pid][_user];
+        userBoost = 1;
+        if (user.veJoeBalance > 0) {
+            userBoost = userBoost.add(user.veJoeBalance.div(pool.totalVeJoe)).mul(maxBoostFactor);
+        }
+    }
+
     /// @notice Harvests JOE from `MASTER_CHEF_V2` MCJV2 and pool `MASTER_PID` to this MCJV3 contract.
```
cryptofish7, line 591: `to this BMCJ contract.`

cryptofish7 (changes requested):
First review, I'm sure there might be some other stuff I'm missing. 

One thing - please adhere all the require reasons strings to a certain format

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+///
+/// This MasterChef also skews how many rewards users receive, it does this by
+/// modifying the algorithm that calculates how many tokens are rewarded to
+/// depositors. Whereas MasterChef calculates rewards based on emission rate and
+/// total liquidity, this version uses adjusted parameters to this calculation.
+///
+/// A users `boostedAmount` (liquidity multiplier) is calculated by the actual supplied
+/// liquidity multiplied by a boost factor. The boost factor is calculated by the
+/// amount of veJOE held by the user over the total veJOE amount held by all pool
+/// participants. total liquidity is the sum of all boosted liquidity.
```
cryptofish7, line 144: `total`->`Total`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+    event Withdraw(address indexed user, uint256 indexed pid, uint256 amount);
+    event UpdatePool(uint256 indexed pid, uint256 lastRewardTimestamp, uint256 lpSupply, uint256 accJoePerShare);
+    event Harvest(address indexed user, uint256 indexed pid, uint256 amount);
+    event EmergencyWithdraw(address indexed user, uint256 indexed pid, uint256 amount);
+    event Init();
+    event MaxBoostFactorChanged(uint256 maxBoostFactor);
+
+    /// @param _MASTER_CHEF_V2 The JoeSwap MCJV2 contract address.
+    /// @param _joe The JOE token contract address.
+    /// @param _MASTER_PID The pool ID of the dummy token on the base MCJV2 contract.
```
cryptofish7, line 231: Missing `@param _veJoe`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        JOE = _joe;
+        VEJOE = _veJoe;
+        MASTER_PID = _MASTER_PID;
+    }
+
+    /// @notice Deposits a dummy token to `MASTER_CHEF_V2` MCJV2. This is required because MCJV2 holds the minting rights for JOE.
+    /// Any balance of transaction sender in `dummyToken` is transferred.
+    /// The allocation point for the pool on MCJV2 is the total allocation point for all pools that receive double incentives.
+    /// @param dummyToken The address of the ERC-20 token to deposit into MCJV2.
+    function init(IERC20 dummyToken) external onlyOwner {
```
cryptofish7, line 248: Rename to `_dummyToken`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        pools = poolInfo.length;
+    }
+
+    /// @notice Add a new LP to the pool. Can only be called by the owner.
+    /// DO NOT add the same LP token more than once. Rewards will be messed up if you do.
+    /// @param allocPoint AP of the new pool.
+    /// @param _lpToken Address of the LP ERC-20 token.
+    /// @param _rewarder Address of the rewarder delegate.
+    function add(
+        uint256 allocPoint,
```
cryptofish7, line 268: Rename to `_allocPoint`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        IERC20 _lpToken,
+        IRewarder _rewarder
+    ) external onlyOwner {
+        require(!lpTokens.contains(address(_lpToken)), "add: LP already added");
+        require(poolInfo.length < 50, "Too many pools");
```
cryptofish7, line 273: Change to `<=`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+    /// @notice Update the given pool's JOE allocation point and `IRewarder` contract. Can only be called by the owner.
+    /// @param _pid The index of the pool. See `poolInfo`.
+    /// @param _allocPoint New AP of the pool.
+    /// @param _rewarder Address of the rewarder delegate.
+    /// @param overwrite True if _rewarder should be `set`. Otherwise `_rewarder` is ignored.
+    function set(
+        uint256 _pid,
+        uint256 _allocPoint,
+        IRewarder _rewarder,
+        bool overwrite
```
cryptofish7, line 308: Change to `_overwrite`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+    function init(IERC20 dummyToken) external onlyOwner {
+        uint256 balance = dummyToken.balanceOf(msg.sender);
+        require(balance != 0, "MasterChefV2: Balance must exceed 0");
+        dummyToken.safeTransferFrom(msg.sender, address(this), balance);
+        dummyToken.approve(address(MASTER_CHEF_V2), balance);
+        MASTER_CHEF_V2.deposit(MASTER_PID, balance);
+        emit Init();
+    }
+
+    /// @notice Returns the number of BMCJ pools.
```
cryptofish7, line 257: Add `@return`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        IRewarder _rewarder,
+        bool overwrite
+    ) external onlyOwner {
+        PoolInfo memory pool = poolInfo[_pid];
+        totalAllocPoint = totalAllocPoint.sub(poolInfo[_pid].allocPoint).add(_allocPoint);
+        pool.allocPoint = _allocPoint;
+        if (overwrite) {
+            _rewarder.onJoeReward(address(0), 0); // sanity check
+            pool.rewarder = _rewarder;
+        }
```
cryptofish7, line 316:
```
        if (overwrite) {
           if (_rewarder != address(0) {
                _rewarder.onJoeReward(address(0), 0); // sanity check
            }
            pool.rewarder = _rewarder;
        }
 ```

File: contracts/BoostedMasterChefJoe.sol [hunk 20]
```diff
@@ -0,0 +1,621 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.6.12;
```
0xnivek, line 3: let's use `0.7.6` to be consistent w/ the other veJOE and sJOE contracts

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+    /// @notice Update reward variables for all pools. Be careful of gas spending!
+    /// @param pids Pool IDs of all to be updated. Make sure to update all active pools.
+    function massUpdatePools(uint256[] calldata pids) external {
+        uint256 len = pids.length;
+        for (uint256 i = 0; i < len; ++i) {
+            updatePool(pids[i]);
+        }
+    }
+
+    /// @notice Calculates and returns the `amount` of JOE per block.
```
cryptofish7, line 380: Add `@return`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        uint256 lpPercent = total.sub(MASTER_CHEF_V2.devPercent()).sub(MASTER_CHEF_V2.treasuryPercent()).sub(
+            MASTER_CHEF_V2.investorPercent()
+        );
+        uint256 lpShare = MASTER_CHEF_V2.joePerSec().mul(lpPercent).div(total);
+        amount = lpShare.mul(MASTER_CHEF_V2.poolInfo(MASTER_PID).allocPoint).div(MASTER_CHEF_V2.totalAllocPoint());
+    }
+
+    /// @notice Update reward variables of the given pool.
+    /// @param pid The index of the pool. See `poolInfo`.
+    function updatePool(uint256 pid) public {
```
cryptofish7, line 392: Change to `_pid`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+            pool.lastRewardTimestamp = block.timestamp;
+            poolInfo[pid] = pool;
+            emit UpdatePool(pid, pool.lastRewardTimestamp, lpSupply, pool.accJoePerShare);
+        }
+    }
+
+    /// @notice Deposit LP tokens to BMCJ for JOE allocation.
+    /// @param pid The index of the pool. See `poolInfo`.
+    /// @param amount LP token amount to deposit.
+    function deposit(uint256 pid, uint256 amount) external nonReentrant {
```
cryptofish7, line 412: Change to `_pid` and `_amount`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+            uint256 joeReward = secondsElapsed.mul(joePerSec()).mul(pool.allocPoint).div(totalAllocPoint);
+            accJoePerShare = accJoePerShare.add(joeReward.mul(ACC_TOKEN_PRECISION).div(pool.totalBoostedAmount));
+        }
+
+        // TODO: Precision.
+        uint256 userBoost = 1000;
+        uint256 userVeJoeBalance = VEJOE.balanceOf(_user);
+        if (userVeJoeBalance > 0) {
+            userBoost = userBoost.add(userVeJoeBalance.mul(maxBoostFactor).div(pool.totalVeJoe));
+        }
```
cryptofish7, line 353: Use `getUserBoost` instead

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        if (address(_rewarder) != address(0)) {
+            _rewarder.onJoeReward(msg.sender, user.amount);
+        }
+        emit Deposit(msg.sender, pid, receivedAmount);
+    }
+
+    /// @notice Withdraw LP tokens from BMCJ.
+    /// @param pid The index of the pool. See `poolInfo`.
+    /// @param amount LP token amount to withdraw.
+    function withdraw(uint256 pid, uint256 amount) external nonReentrant {
```
cryptofish7, line 475: Change to `_pid` and `_amount`

contracts/BoostedMasterChefJoe.sol
0xnivek, line 144: nit: `total` -> `Total`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+
+            // Increase claimableJoe
+            claimableJoe[pid][_user] += pending;
+
+            // Update pool.totalVeJoe
+            pool.totalVeJoe = pool.totalVeJoe.sub(user.veJoeBalance).add(_newUserBalance);
+
+            // Decrement the total boosted amount and increment it after updating the
+            // users veJoeBalance .
+            pool.totalBoostedAmount = pool.totalBoostedAmount.sub(user.amount.mul(getUserBoost(pid, _user)));
```
cryptofish7, line 568: Need to divide by 1000

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+
+            // Update users veJoeBalance.
+            user.veJoeBalance = _newUserBalance;
+
+            // Update the pools total boosted amount.
+            pool.totalBoostedAmount = pool.totalBoostedAmount.add(user.amount.mul(getUserBoost(pid, _user)));
```
cryptofish7, line 574: Need to divide by 1000

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+    /// @notice Returns a users boosted emission share
+    /// @param _pid The index of the pool. See `poolInfo`
+    /// @param _user The address of the depositor
+    /// @dev `userBoost` is scaled by 1e3
+    function getUserBoost(uint256 _pid, address _user) public view returns (uint256 userBoost) {
+        PoolInfo memory pool = poolInfo[_pid];
+        UserInfo memory user = userInfo[_pid][_user];
+        userBoost = 1000;
+        if (user.veJoeBalance > 0 && pool.totalVeJoe > 0) {
+            userBoost = userBoost.add(user.veJoeBalance.mul(maxBoostFactor).div(pool.totalVeJoe));
```
cryptofish7, line 594: Why not divide by 1000 here instead?

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+                .mul(pool.accJoePerShare)
+                .div(ACC_TOKEN_PRECISION)
+                .div(1000);
+        }
+    }
+
+    /// @notice Returns a users boosted emission share
+    /// @param _pid The index of the pool. See `poolInfo`
+    /// @param _user The address of the depositor
+    /// @dev `userBoost` is scaled by 1e3
```
cryptofish7, line 588: Missing `@return`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+    }
+
+    /// @notice Harvests JOE from `MASTER_CHEF_V2` MCJV2 and pool `MASTER_PID` to this BMCJ contract.
+    function harvestFromMasterChef() public {
+        MASTER_CHEF_V2.deposit(MASTER_PID, 0);
+    }
+
+    /// @notice Withdraw without caring about rewards. EMERGENCY ONLY.
+    /// @param pid The index of the pool. See `poolInfo`.
+    function emergencyWithdraw(uint256 pid) external nonReentrant {
```
cryptofish7, line 605: Change to `_pid`

cryptofish7 (changes requested):

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        require(balance != 0, "MasterChefV2: Balance must exceed 0");
+        dummyToken.safeTransferFrom(msg.sender, address(this), balance);
+        dummyToken.approve(address(MASTER_CHEF_V2), balance);
+        MASTER_CHEF_V2.deposit(MASTER_PID, balance);
+        emit Init();
+    }
+
+    /// @notice Returns the number of BMCJ pools.
+    function poolLength() external view returns (uint256 pools) {
+        pools = poolInfo.length;
```
0xnivek, line 259: nit: `numPools` might be a better naming

contracts/BoostedMasterChefJoe.sol
0xnivek, line 273: for the error messaging, I believe the convention we've been using is `<contract name: error message`, e.g. `BoostedMasterChefJoe: LP already added`

0xnivek, line 273: do we want to make the max number of pools configurable?

contracts/BoostedMasterChefJoe.sol
0xnivek, line 268: nit: `_allocPoint`

contracts/BoostedMasterChefJoe.sol
0xnivek, line 308: nit: `_overwrite`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+    /// @param _rewarder Address of the rewarder delegate.
+    /// @param overwrite True if _rewarder should be `set`. Otherwise `_rewarder` is ignored.
+    function set(
+        uint256 _pid,
+        uint256 _allocPoint,
+        IRewarder _rewarder,
+        bool overwrite
+    ) external onlyOwner {
+        PoolInfo memory pool = poolInfo[_pid];
+        totalAllocPoint = totalAllocPoint.sub(poolInfo[_pid].allocPoint).add(_allocPoint);
```
0xnivek, line 311: nit: `poolInfo[_pid].allocPoint` -> `pool.allocPoint`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        uint256 userBoost = 1000;
+        uint256 userVeJoeBalance = VEJOE.balanceOf(_user);
+        if (userVeJoeBalance > 0) {
+            userBoost = userBoost.add(userVeJoeBalance.mul(maxBoostFactor).div(pool.totalVeJoe));
+        }
+        pendingJoe = user
+            .amount
+            .mul(userBoost)
+            .mul(accJoePerShare)
+            .div(1000)
```
0xnivek, line 358: is this `1000` coming from `userBoost`, if so let's replace this with `userBoost` instead of hardcoding the `1000`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        if (address(pool.rewarder) != address(0)) {
+            bonusTokenAddress = address(pool.rewarder.rewardToken());
+            bonusTokenSymbol = IERC20(pool.rewarder.rewardToken()).safeSymbol();
+            pendingBonusToken = pool.rewarder.pendingTokens(_user);
+        }
+    }
+
+    /// @notice Update reward variables for all pools. Be careful of gas spending!
+    /// @param pids Pool IDs of all to be updated. Make sure to update all active pools.
+    function massUpdatePools(uint256[] calldata pids) external {
```
0xnivek, line 373: nit: `_pids`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        uint256 len = pids.length;
+        for (uint256 i = 0; i < len; ++i) {
+            updatePool(pids[i]);
+        }
+    }
+
+    /// @notice Calculates and returns the `amount` of JOE per block.
+    function joePerSec() public view returns (uint256 amount) {
```
0xnivek, line 381: nit: `_amount`

contracts/BoostedMasterChefJoe.sol
0xnivek, line 392: nit: `_pid`

contracts/BoostedMasterChefJoe.sol
0xnivek, line 412: nit: `_pid` and `_amount`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+    /// @param amount LP token amount to deposit.
+    function deposit(uint256 pid, uint256 amount) external nonReentrant {
+        harvestFromMasterChef();
+        updatePool(pid);
+        PoolInfo storage pool = poolInfo[pid];
+        UserInfo storage user = userInfo[pid][msg.sender];
+
+        if (user.amount > 0) {
+            // Harvest JOE
+            uint256 boostedAmount = user.amount.mul(getUserBoost(pid, msg.sender)).div(1000);
```
0xnivek, line 420: same question regarding where the `1000` is coming from, should we store it as a variable so we can re-use it across the contract?

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+                claimableJoe[pid][msg.sender]
+            );
+            claimableJoe[pid][msg.sender] = 0;
+            JOE.safeTransfer(msg.sender, pending);
+            emit Harvest(msg.sender, pid, pending);
+        }
+
+        uint256 balanceBefore = pool.lpToken.balanceOf(address(this));
+        pool.lpToken.safeTransferFrom(msg.sender, address(this), amount);
+        uint256 receivedAmount = pool.lpToken.balanceOf(address(this)).sub(balanceBefore);
```
0xnivek, line 432: is there a case in which `receivedAmount != amount`?

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        // Effects
+
+        /// If a user already has an amount we need to update the `poolInfo`
+        /// `totalBoostedAmount` and `totalVeJoe` variables relative to their
+        /// old amount. Otherwise we can just add the new amounts onto the
+        /// existing.
+
+        // Update the `pool.totalVeJoe`.
+        pool.totalVeJoe = pool.totalVeJoe.sub(user.veJoeBalance).add(VEJOE.balanceOf(msg.sender));
+        user.veJoeBalance = VEJOE.balanceOf(msg.sender);
```
0xnivek, line 443: two thoughts here regarding potentially saving some gas:
1. Maybe we can check if `user.veJoeBalance != VEJOE.balanceOf(msg.sender)` and only perform the update to `pool.totalVeJoe` if that is the case
2. We can store `VEJOE.balanceOf(msg.sender)` into a local variable so that we don't have to call it twice

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+            pool.totalBoostedAmount = pool.totalBoostedAmount.sub(user.amount.mul(userBoost).div(1000)).add(
+                user.amount.add(amount).mul(userBoost).div(1000)
+            );
+        }
+
+        user.amount = user.amount.add(receivedAmount);
+        user.rewardDebt = user
+            .amount
+            .sub(amount)
+            .mul(getUserBoost(pid, msg.sender))
```
0xnivek, line 459: can we reuse the `userBoost` on line 446 or is this called again intentionally here?

contracts/BoostedMasterChefJoe.sol
0xnivek, line 475: nit: `_pid` and `_amount`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,621 @@
...
+        // veJoe balance from `pool.totalVeJoe`, and full boosted balance
+        // from `pool.totalBoostedAmount` otherwise we just adjust down the
+        // total boosted amount as total vejoe is unchanged.
+
+        // Update the `pool.totalBoostedAmount` and `totalVeJoe` if needed.
+        if (user.amount.sub(amount) == 0) {
+            pool.totalVeJoe = pool.totalVeJoe.sub(VEJOE.balanceOf(msg.sender));
+            pool.totalBoostedAmount = pool.totalBoostedAmount.sub(user.amount.mul(userBoost).div(1000));
+        } else {
+            pool.totalBoostedAmount = pool.totalBoostedAmount.sub(amount.mul(userBoost).div(1000));
```
0xnivek, line 508: i think we can extract these two statements to outside of the if block, i.e. just do:
```
if (user.amount.sub(amount) == 0) { ... }
pool.totalBoostedAmount = pool.totalBoostedAmount.sub(amount.mul(userBoost).div(1000));
```

0xnivek (commented):
awesome work!!

contracts/BoostedMasterChefJoe.sol
jummy123: The only issue I see with that is that I started with a base of MasterChefJoeV3 and have just changed some methods but left all the other logic unchanged. Will need to research that the existing part of the contract still works with the updated version.

contracts/BoostedMasterChefJoe.sol
jummy123: Hm looking at the existing code a complete mixture of conventions were used, I'm going to change everything to use the recent convention we have been using.

jummy123: We are limited by gas here and ATM 50 is just a figure I was told, I need to do proper testing to find out a real max number, after that not sure there is any point in making it configurable.

contracts/BoostedMasterChefJoe.sol
jummy123: I'm going to leave this change until later, this code is vebatim from MCV3 and I'm hesitant to change it.

contracts/BoostedMasterChefJoe.sol
jummy123: I'm going to leave this change until later, this code is vebatim from MCV3 and I'm hesitant to change it.

contracts/BoostedMasterChefJoe.sol
jummy123: I've moved this into being needed in just one function (I don't think I need to make it a named variable now). It is there because the `maxBoostFactor` is 1500 (which represents 1.5%).

contracts/BoostedMasterChefJoe.sol
jummy123: see comment above

contracts/BoostedMasterChefJoe.sol
jummy123: The only one I can think of is reflection tokens, which may not be applicable to uniswap pairs.

contracts/BoostedMasterChefJoe.sol
jummy123: Have removed this altogether as all veJoe balance changes should now be updated by the token callback function.

contracts/BoostedMasterChefJoe.sol
jummy123: :+1: this is fixed now

contracts/BoostedMasterChefJoe.sol
jummy123: The reason for that is:
example:
```
userBoost = 1000
user.veJoeBalance = 10
pool.totalVeJoe = 10
maxBoostFactor = 1500
1000 + (10 * 1500 / 10)
1000 + (1500)
2500
```
If I divide this by 1000 here the value will round to `2` where really it should be `2.5`.

contracts/BoostedMasterChef.sol
jummy123: It should but I have changed to do the addition first.

cryptofish7:
> I started this code by copying `MasterChefJoeV3` and modifying the necessary functions for boosted staking. I know it is not clear from the diff but a lot of the code is copied verbatim from MCJV3. This means we have a couple of choices to make with tradeoffs on both sides.
> 
> * Changing the solidity version
> * Refactoring/renaming unmodified code from MCJV3
> * Rewrite/cleanup all the copied natspec (puncutation etc)
> 
> I'm erring on the side of renaming variables and error messages to our standard conventions but not modifying the solidity version or unchanged functions. Open for discussion 😵‍💫

I agree with you here. I wrote MCJV3 and the code is not great, so we should take the opportunity to polish it up.

As for the version, would prefer to avoid having multiple versions in the same repo cos it gets annoying when trying to import OZ dependencies. Original version was 0.6.12, but we also added 0.7.6 for StableJoeStaking -- I would defer to one of those two versions.

contracts/BoostedMasterChefJoe.sol
cryptofish7: Ah yes good point

contracts/BoostedMasterChefJoe.sol
cryptofish7: I think either should be fine here. Big question - do we/should we make it upgradeable?

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,587 @@
...
+    event Set(uint256 indexed pid, uint256 allocPoint, IRewarder indexed rewarder, bool overwrite);
+    event Deposit(address indexed user, uint256 indexed pid, uint256 amount);
+    event Withdraw(address indexed user, uint256 indexed pid, uint256 amount);
+    event UpdatePool(uint256 indexed pid, uint256 lastRewardTimestamp, uint256 lpSupply, uint256 accJoePerShare);
+    event Harvest(address indexed user, uint256 indexed pid, uint256 amount);
+    event EmergencyWithdraw(address indexed user, uint256 indexed pid, uint256 amount);
+    event Init();
+    event MaxBoostFactorChanged(uint256 maxBoostFactor);
+
+    /// @param _MASTER_CHEF_V2 The JoeSwap MCJV2 contract address.
```
cryptofish7, line 221: `The MCJV2 contract address.`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,587 @@
...
+            UserInfo storage user = userInfo[pid][_user];
+
+            // Skip if user doesn't have any deposit in the pool
+            if (user.amount == 0) {
+                continue;
+            }
+
+            PoolInfo storage pool = poolInfo[pid];
+            updatePool(pid);
+            // calculate pending
```
cryptofish7, line 515: `calculate`->`Calculate`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,587 @@
...
+            uint256 oldBoostedLiquidity = getBoostedLiquidity(pid, _user);
+            uint256 pending = oldBoostedLiquidity.mul(pool.accJoePerShare).div(ACC_TOKEN_PRECISION).sub(
+                user.rewardDebt
+            );
+
+            // Increase claimableJoe
+            claimableJoe[pid][_user] += pending;
```
cryptofish7, line 522: Since we're not using 0.8 or higher, change to `claimableJoe[pid][_user] = claimable[pid][_user].add(pending)`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,587 @@
...
+    /// @notice Returns a users boosted emission share
+    /// @param _pid The index of the pool. See `poolInfo`
+    /// @param _user The address of the depositor
+    /// @dev `userBoost` is scaled by 1e3
+    /// @return userBoost A multiplier for a users liquidity
+    function getUserBoost(uint256 _pid, address _user) public view returns (uint256 userBoost) {
+        PoolInfo memory pool = poolInfo[_pid];
+        UserInfo memory user = userInfo[_pid][_user];
+        userBoost = 1000;
+        if (user.veJoeBalance > 0 && pool.totalVeJoe > 0) {
```
cryptofish7, line 559: Can do `if (user.veJoeBalance != 0 && pool.totalVeJoe != 0)` instead to save gas

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,587 @@
...
+    /// @notice Withdraw LP tokens from BMCJ.
+    /// @param _pid The index of the pool. See `poolInfo`.
+    /// @param _amount LP token amount to withdraw.
+    function withdraw(uint256 _pid, uint256 _amount) external nonReentrant {
+        harvestFromMasterChef();
+        updatePool(_pid);
+        PoolInfo storage pool = poolInfo[_pid];
+        UserInfo storage user = userInfo[_pid][msg.sender];
+
+        if (user.amount > 0) {
```
cryptofish7, line 458: `if (user.amount != 0)` to save gas

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,587 @@
...
+
+        uint256 balanceBefore = pool.lpToken.balanceOf(address(this));
+        pool.lpToken.safeTransferFrom(msg.sender, address(this), _amount);
+        uint256 receivedAmount = pool.lpToken.balanceOf(address(this)).sub(balanceBefore);
+
+        // Effects
+
+        // Update the `pool.totalVeJoe`
+        uint256 veJoeBalance = VEJOE.balanceOf(msg.sender);
+        if (veJoeBalance > 0) {
```
cryptofish7, line 426: `if (veJoeBalance != 0)` to save gas

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,587 @@
...
+    /// @param _pid The index of the pool. See `poolInfo`.
+    /// @param _amount LP token amount to deposit.
+    function deposit(uint256 _pid, uint256 _amount) external nonReentrant {
+        harvestFromMasterChef();
+        updatePool(_pid);
+        PoolInfo storage pool = poolInfo[_pid];
+        UserInfo storage user = userInfo[_pid][msg.sender];
+
+        // Pay a user any pending rewards
+        if (user.amount > 0) {
```
cryptofish7, line 404: `if (user.amount != 0)` to save gas

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,587 @@
...
+        _amount = lpShare.mul(MASTER_CHEF_V2.poolInfo(MASTER_PID).allocPoint).div(MASTER_CHEF_V2.totalAllocPoint());
+    }
+
+    /// @notice Update reward variables of the given pool.
+    /// @param _pid The index of the pool. See `poolInfo`.
+    function updatePool(uint256 _pid) public {
+        PoolInfo memory pool = poolInfo[_pid];
+        if (block.timestamp > pool.lastRewardTimestamp) {
+            uint256 lpSupply = pool.lpToken.balanceOf(address(this));
+            if (lpSupply > 0) {
```
cryptofish7, line 381: `if (lpSupply != 0)` to save gas

cryptofish7 (changes requested):

contracts/BoostedMasterChefJoe.sol
jummy123: Fixed this (and also a related bug in `updateBoost`) and added a test for claimable.

contracts/BoostedMasterChefJoe.sol
jummy123: I'm tempted to leave this as is, I greped through the codebase and their seems to be some convention of having a `poolLength` function, possibly used for frontend.

contracts/VeJoeToken.sol
```diff
@@ -6,10 +6,17 @@ import "@openzeppelin/contracts/access/Ownable.sol";
...
 
+interface IBoostedMasterChefJoe {
+    function updateBoost(address, uint256) external;
+}
+
 /// @title Vote Escrow Joe Token - veJOE
 /// @author Trader Joe
 /// @notice Infinite supply, used to receive extra farming yields and voting power
 contract VeJoeToken is VeERC20("VeJoeToken", "veJOE"), Ownable {
+    /// @notice the masterPlatypus contract
```
0xnivek, line 17: nit: `masterPlatypus` -> `BoostedMasterChefJoe`

test/BoostedMasterChef.test.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    // Transfer some vejoe to bob
+    await this.veJoe.connect(this.dev).mint(this.bob.address, 100)
+
+    // Bob enters the pool
+    await this.lp.connect(this.bob).approve(this.bmc.address, "1000")
+    await this.bmc.connect(this.bob).deposit(0, "1000")
+    pool = await this.bmc.poolInfo(0)
+    expect(pool.totalVeJoe).to.equal(100)
+    expect(pool.totalBoostedAmount).to.equal("3500")
```
0xnivek, line 91: any reason these are strings instead of just numbers? (i.e. `3500` instead of `"3500"`)

0xnivek (commented):
🔥

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,589 @@
...
+
+interface IRewarder {
+    function onJoeReward(address user, uint256 newLpAmount) external;
+
+    function pendingTokens(address user) external view returns (uint256 pending);
+
+    function rewardToken() external view returns (IERC20);
+}
+
+library BoringERC20 {
```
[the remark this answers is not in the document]
jummy123: Yeah this is a great call, I have pushed these changes. Annoyingly there are already interfaces there but because of dependency issues I had to basically duplicate the files and change the names. It's not a great solution but at least this file is cleaned up now.

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,589 @@
...
+        VEJOE = _veJoe;
+        MASTER_PID = _MASTER_PID;
+    }
+
+    /// @notice Deposits a dummy token to `MASTER_CHEF_V2` MCJV2. This is required because MCJV2
+    /// holds the minting rights for JOE.  Any balance of transaction sender in `_dummyToken` is transferred.
+    /// The allocation point for the pool on MCJV2 is the total allocation point for all pools that receive
+    /// double incentives.
+    /// @param _dummyToken The address of the ERC-20 token to deposit into MCJV2.
+    function init(IERC20 _dummyToken) external onlyOwner {
```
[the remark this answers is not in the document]
jummy123: I'm not actually sure TBH, this is the funding mechanism for all the pools. @cryptofish7 does this method ever need to be called more than once?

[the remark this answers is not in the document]
cryptofish7: Technically no. Let's add an `initialized` variable so it coincides with the function name.

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,474 @@
...
+        uint256 _pid,
+        uint256 _allocPoint,
+        IRewarder _rewarder,
+        bool _overwrite
+    ) external onlyOwner {
+        PoolInfo memory pool = poolInfo[_pid];
+        totalAllocPoint = totalAllocPoint.sub(poolInfo[_pid].allocPoint).add(_allocPoint);
+        pool.allocPoint = _allocPoint;
+        if (_overwrite) {
+            _rewarder.onJoeReward(address(0), 0); // sanity check
```
cryptofish7, line 195: Change to:
```
if (_overwrite) {
   if (address(_rewarder) != address(0)) {
      _rewarder.onJoeReward(address(0), 0);
   }
   pool.rewarder = _rewarder;
}
```

Reason is because we want to have the option to clear a rewarder.

cryptofish7 (changes requested):

contracts/BoostedMasterChefJoe.sol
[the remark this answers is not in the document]
jummy123: I think I have solved this in a _smart_ way without using an extra state variable. As we cannot change the value of `MASTER_PID` variable after contract instantiation it is enough to check that the address of the current contract doesn't have a balance in that pool. Fix pushed with test.

test/BoostedMasterChef.test.ts
jummy123: Definitely not a good reason, I will either change these to be homogeneous or use `ethers.utils.parseEther`

contracts/BoostedMasterChefJoe.sol
cryptofish7: No reason other than the original MasterChef wasnt upgradeable.

contracts/BoostedMasterChefJoe.sol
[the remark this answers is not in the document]
cryptofish7: Nice idea. Wondering if we should be more strict and make it `MASTER_CHEF_V2.poolInfo(MASTER_PID).amount == 0` instead

test/BoostedMasterChef.test.ts
```diff
@@ -0,0 +1,348 @@
...
+    await network.provider.send("evm_setAutomine", [true])
+  })
+
+  it("should record the correct reward debt on withdraw", async function () {
+    await this.veJoe.connect(this.dev).mint(this.bob.address, 100)
+    await this.lp.connect(this.bob).approve(this.bmc.address, 1000)
+
+    await this.bmc.connect(this.bob).deposit(0, 1000)
+    await network.provider.send("evm_setAutomine", [false])
+    // Make sure contract has to JOE to emit
```
cryptofish7, line 193: grammar typo

test/BoostedMasterChef.test.ts
```diff
@@ -0,0 +1,348 @@
...
+    await this.bmc.connect(this.dev).harvestFromMasterChef()
+    await network.provider.send("evm_increaseTime", [3600])
+    await network.provider.send("evm_mine")
```
cryptofish7, line 196: Can use `increase(duration.hours(1))` from test/utilities/time.ts instead

test/BoostedMasterChef.test.ts
```diff
@@ -0,0 +1,348 @@
...
+
+    await this.bmc.connect(this.bob).withdraw(0, 0)
+
+    await network.provider.send("evm_increaseTime", [1])
+    await network.provider.send("evm_mine")
```
cryptofish7, line 201: Can use `increase(duration.seconds(1))` from test/utilities/time.ts instead

test/BoostedMasterChef.test.ts
```diff
@@ -0,0 +1,348 @@
...
+    await network.provider.send("evm_setAutomine", [true])
+  })
+
+  it("should claim reward on deposit", async function () {
+    await this.veJoe.connect(this.dev).mint(this.bob.address, 100)
+    await this.lp.connect(this.bob).approve(this.bmc.address, 1000)
+
+    await this.bmc.connect(this.bob).deposit(0, 500)
+    await network.provider.send("evm_setAutomine", [false])
+    // Make sure contract has to JOE to emit
```
cryptofish7, line 215: Grammar typo

test/BoostedMasterChef.test.ts
```diff
@@ -0,0 +1,348 @@
...
+    await this.bmc.connect(this.dev).harvestFromMasterChef()
+    await network.provider.send("evm_increaseTime", [3600])
+    await network.provider.send("evm_mine")
```
cryptofish7, line 218: Can use `increase(duration.hours(1))` from test/utilities/time.ts instead

test/BoostedMasterChef.test.ts
```diff
@@ -0,0 +1,348 @@
...
+  })
+
+  it("should stop boosting if burn vejoe", async function () {
+    // Bob enters the pool
+    await this.veJoe.connect(this.dev).mint(this.bob.address, 100)
+    await this.lp.connect(this.bob).approve(this.bmc.address, 1000)
+    await this.bmc.connect(this.bob).deposit(0, 1000)
+
+    await network.provider.send("evm_increaseTime", [3600])
+    await network.provider.send("evm_mine")
```
cryptofish7, line 308: Can use `increase(duration.hours(1))` from test/utilities/time.ts instead

test/BoostedMasterChef.test.ts
```diff
@@ -0,0 +1,348 @@
...
+    await this.veJoe.connect(this.dev).mint(this.bob.address, 100)
+    await this.lp.connect(this.bob).approve(this.bmc.address, 1000)
+    await this.lp.connect(this.alice).approve(this.bmc.address, 1000)
+
+    await this.bmc.connect(this.bob).deposit(0, 1000)
+    await this.bmc.connect(this.alice).deposit(0, 1000)
+    await network.provider.send("evm_mine")
+
+    await network.provider.send("evm_increaseTime", [3600])
+    await network.provider.send("evm_mine")
```
cryptofish7, line 332: Can use `increase(duration.hours(1))` from test/utilities/time.ts instead

cryptofish7 (approved):

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,480 @@
...
+    /// @dev Total allocation points. Must be the sum of all allocation points in all pools
+    uint256 public totalAllocPoint;
+    uint256 private constant ACC_TOKEN_PRECISION = 1e18;
+    /// @dev A maximum scaling factor applied to boosted amounts
+    uint256 public maxBoostFactor = 1500;
+    /// @dev Amount of claimable Joe the user has, this is required as we
+    /// need to update rewardDebt after a token operation but we don't
+    /// want to send a reward at this point. This amount gets added onto
+    /// the pending amount when a user claims
+    mapping(uint256 => mapping(address => uint256)) public claimableJoe;
```
0xnivek, line 95: nit: would be nice to add some spacing in between vars to make it easier to read

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,480 @@
...
+    IMasterChefJoe public immutable MASTER_CHEF_V2;
+    /// @notice Address of JOE contract
+    IERC20 public immutable JOE;
+    /// @notice Address of veJOE contract
+    IERC20 public immutable VEJOE;
+    /// @notice The index of BMCJ master pool in MCJV2
+    uint256 public immutable MASTER_PID;
+    /// @notice Info of each BMCJ pool
+    PoolInfo[] public poolInfo;
+    // Set of all LP tokens that have been added as pools
```
0xnivek, line 82: mark as `@dev`

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,480 @@
...
+    function massUpdatePools(uint256[] calldata _pids) external {
+        uint256 len = _pids.length;
+        for (uint256 i = 0; i < len; ++i) {
+            updatePool(_pids[i]);
+        }
+    }
+
+    /// @notice Calculates and returns the `amount` of JOE per second
+    /// @return _amount The amount of JOE emitted per second
+    function joePerSec() public view returns (uint256 _amount) {
```
0xnivek, line 259: question: whats our convention on the naming of return variables? asking b/c this return has an underscore but the ones in `pendingTokens` function doesn't so we should probably be consistent

0xnivek (commented):
⭐

contracts/BoostedMasterChefJoe.sol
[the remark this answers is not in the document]
jummy123: Good idea, there is no `amount` variable on `PoolInfo` structs so changed this to `_dummyToken.balanceOf(address(MASTER_CHEF_V2)) == 0`

contracts/BoostedMasterChefJoe.sol
jummy123: Hmm, this is the only function that prefixes the return name with `_`. For the time being I have removed this to make consistent with the rest of the functions. The solidity style guide doesn't mention anything about this but there is an unofficial convention of prefixing local variables with `_` or using it when there would be a name clash with a state variable.

I'm totally up for writing our own style guide in notion for things not covered by the solidity style guide.

contracts/BoostedMasterChefJoe.sol
```diff
@@ -0,0 +1,480 @@
...
+
+    event Add(uint256 indexed pid, uint256 allocPoint, IERC20 indexed lpToken, IRewarder indexed rewarder);
+    event Set(uint256 indexed pid, uint256 allocPoint, IRewarder indexed rewarder, bool overwrite);
+    event Deposit(address indexed user, uint256 indexed pid, uint256 amount);
+    event Withdraw(address indexed user, uint256 indexed pid, uint256 amount);
+    event UpdatePool(uint256 indexed pid, uint256 lastRewardTimestamp, uint256 lpSupply, uint256 accJoePerShare);
+    event Harvest(address indexed user, uint256 indexed pid, uint256 amount);
+    event EmergencyWithdraw(address indexed user, uint256 indexed pid, uint256 amount);
+    event Init();
+    event MaxBoostFactorChanged(uint256 maxBoostFactor);
```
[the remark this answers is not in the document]
jummy123: Yeah I don't think it is possible to change maxBoostFactor as it would require iterating over every user in the pool to update their boostedLiquidity, will remove this event.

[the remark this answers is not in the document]
jummy123: I don't think we can use immutable/constant now as I made the contracts upgrade able.

contracts/BoostedMasterChefJoe.sol
jummy123: I have changed this to use upgradable proxy.

contracts/BoostedMasterChefJoe.sol
jummy123: Inclined to leave this in as it covers both cases.

LouisMeMyself (approved):

contracts/BoostedMasterChefJoe.sol
cryptofish7: This part is safe to change

cryptofish7 (changes requested):

contracts/BoostedMasterChefJoe.sol
jummy123: Should I need to change the line above this to `PoolInfo storage pool = poolInfo[_pid];`?

jummy123: Fixed and test added.

cryptofish7 (approved):
LGTM

cryptofish7 (approved):

PR #39, opened 2022-02-12
Handle veJoeStaking edge cases + minor cleanups

0xnivek:
There are some edge cases in `VeJoeStaking` that I realized I didn't handle in my initial PR. Specifically:

1. When the user deposits `boostedThreshold` amount of `JOE`, we should extend their boost period **even if they are currently already receiving boosted benefits**.
2. We shouldn't allow passive accrual of veJOE after the user has hit their max veJOE cap. For example, imagine the following scenario:
  a. User has some amount of staked JOE and max veJOE cap.
  b. After `s` seconds, the user deposits some more JOE — this means that their max veJOE cap now increases! However, the user should not be accruing any veJOE for those `s` seconds where they were at their max veJOE cap. We can make sure of this by updating the user's `lastRewardTimestamp` inside `deposit` if they are at their max veJOE cap.

I've added tests as well for these two edge cases.

### DIFF
--- a/contracts/VeJoeStaking.sol
+++ b/contracts/VeJoeStaking.sol
@@ -179,6 +179,16 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
 
             uint256 userStakedJoe = userInfo.balance;
 
+            uint256 userVeJoeBalance = veJoe.balanceOf(_user);
+            uint256 userMaxVeJoeCap = userStakedJoe.mul(maxCap);
+
+            // If the user is currently at their max veJOE cap, we need to update
+            // their `lastRewardTimestamp` to now to prevent passive veJOE accrual
+            // after hitting their max cap.
+            if (userVeJoeBalance >= userMaxVeJoeCap) {
+                userInfo.lastRewardTimestamp = block.timestamp;
+            }
+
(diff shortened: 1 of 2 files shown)


LouisMeMyself (approved):
Nice edge case, lgtm

contracts/VeJoeStaking.sol
```diff
@@ -173,12 +180,21 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
+                userInfo.lastRewardTimestamp = block.timestamp;
+            }
+
             userInfo.balance = userStakedJoe.add(_amount);
 
-            // User is eligible for boosted benefits if and only if all of the following are true:
-            // - User is not already currently receiving boosted benefits
-            // - `_amount` is at least `boostedThreshold / 100 * userStakedJoe`
-            if (userInfo.boostEndTimestamp == 0 && _amount.mul(100) >= boostedThreshold.mul(userStakedJoe)) {
+            // User is eligible for boosted benefits `_amount` is at least
```
cryptofish7, line 195: `is`->`if`

cryptofish7 (changes requested):

cryptofish7 (approved):

PR #40, opened 2022-02-12
Add checks for certain variables in veJOE staking

0xnivek:
This PR adds checks to enforce:

- `boostedThreshold > 0`
- `boostedDuration <= 365 days`
- `maxCap <= 100000`

and also adds tests

### DIFF
--- a/contracts/VeJoeStaking.sol
+++ b/contracts/VeJoeStaking.sol
@@ -106,7 +106,7 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
             _boostedThreshold != 0 && _boostedThreshold <= 100,
             "VeJoeStaking: expected _boostedThreshold to be > 0 and <= 100"
         );
-        require(_boostedDuration <= MAX_BOOSTED_DURATION, "VeJoeStaking: expected _boostedDuration to be <= 15 days");
+        require(_boostedDuration <= MAX_BOOSTED_DURATION, "VeJoeStaking: expected _boostedDuration to be <= 365 days");
         require(
             _maxCap != 0 && _maxCap <= MAX_MAX_CAP,
             "VeJoeStaking: expected new _maxCap to be non-zero and <= 100000"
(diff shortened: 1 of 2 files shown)


contracts/VeJoeStaking.sol
```diff
@@ -41,6 +41,9 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
     /// For example, if user has `n` JOE staked, they can own a maximum of `n * maxCap` veJOE.
     uint256 public maxCap;
 
+    /// @notice The upper limit of `maxCap`
+    uint256 public MAX_MAX_CAP;
```
LouisMeMyself, line 45: I don’t think this is a good name for a variable..
Maybe `UPPER_BOUND_MAX_CAP` or `UPPER_LIMIT_MAX_CAP` ?

contracts/VeJoeStaking.sol
```diff
@@ -60,6 +63,9 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
     /// @notice The length of time a user receives boosted benefits
     uint256 public boostedDuration;
 
+    /// @notice The upper limit of `boostedDuration`
+    uint256 public MAX_BOOSTED_DURATION;
```
LouisMeMyself, line 67: If we change the name for the previous upper bound, we should rename this too

contracts/VeJoeStaking.sol
```diff
@@ -60,6 +63,9 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
     /// @notice The length of time a user receives boosted benefits
     uint256 public boostedDuration;
 
+    /// @notice The upper limit of `boostedDuration`
+    uint256 public UPPER_LIMIT_BOOSTED_DURATION;
```
jummy123, line 67: One of the feedback I have received before is not to name these variables in UPPERCASE_WITH_UNDERSCORES as they aren't actually `immutable` variables.

contracts/VeJoeStaking.sol
```diff
@@ -87,17 +93,26 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
         uint256 _boostedDuration,
         uint256 _maxCap
     ) public initializer {
+        UPPER_LIMIT_BOOSTED_DURATION = 365 days;
```
jummy123, line 96: Optional, you can define this just before the `require` statements where they are used, saving gas on failed txs.

contracts/VeJoeStaking.sol
```diff
@@ -87,17 +93,26 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
+        UPPER_LIMIT_MAX_CAP = 100000;
```
cryptofish7, line 97: Can we define `UPPER_LIMIT_MAX_CAP`, `UPPER_LIMIT_BOOSTED_DURATION` and `PRECISION` as immutable variables and assign them in a constructor instead?

Or, we don't use upper snake case.

cryptofish7 (changes requested):

cryptofish7 (approved):

PR #41, opened 2022-02-17
Simpler vejoe staking

LouisMeMyself:


contracts/VeJoeStaking.sol
```diff
@@ -244,14 +145,14 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
             userInfo.balance >= _amount,
             "VeJoeStaking: cannot withdraw greater amount of JOE than currently staked"
         );
+        update();
+        //We don't need to claim as veJOE balance will be reseted to 0
```
jummy123, line 149: tiny nit, there should be a space between `//` and `We`

contracts/VeJoeStaking.sol
```diff
@@ -44,71 +37,43 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
-    /// The only exception is the user will also receive boosted benefits if it's their first
-    /// time staking.
-    uint256 public boostedThreshold;
+    /// @notice The last time that the reward variables were updated
+    uint256 public lastRewardTimestamp;
 
-    /// @notice The length of time a user receives boosted benefits
-    uint256 public boostedDuration;
+    /// @notice Rate of veJOE generated per sec per JOE staked, scaled to `GENERATION_RATE_PRECISION`
+    uint256 public baseGenerationRate;
```
cryptofish7, line 50: If we're removing `boostedGenerationRate`, is there any point to having `baseGenerationRate`?

contracts/VeJoeStaking.sol
```diff
@@ -245,13 +147,14 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
             "VeJoeStaking: cannot withdraw greater amount of JOE than currently staked"
         );
 
+        updateRewardVars();
+        // Note that we don't need to claim as the user's veJOE balance will be reset to 0
+
         userInfo.balance = userInfo.balance.sub(_amount);
```
cryptofish7, line 153:
```
        updateRewardVars();

        // Note that we don't need to claim as the user's veJOE balance will be reset to 0
        userInfo.balance = userInfo.balance.sub(_amount);
```

contracts/VeJoeStaking.sol
```diff
@@ -262,98 +165,65 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
+        updateRewardVars();
         _claim();
     }
 
     /// @notice Get the pending amount of veJOE for a given user
     /// @param _user The user to lookup
-    /// @return The number of pending veJOE tokens for `_user`
-    function getPendingVeJoe(address _user) public view returns (uint256) {
+    /// @return pendingVeJoe The number of pending veJOE tokens for `_user`
+    function getPendingVeJoe(address _user) public view returns (uint256 pendingVeJoe) {
```
cryptofish7, line 175: Any reason why you choose to declare the return variable here?

cryptofish7 (changes requested):

contracts/VeJoeStaking.sol
LouisMeMyself: Indeed better not to, as we don't always return `pendingVeJoe`

contracts/VeJoeStaking.sol
cryptofish7: Ah yes good point. In that case, let's rename to `veJoePerSec`

cryptofish7 (changes requested):

cryptofish7 (approved):

PR #43, opened 2022-02-19
Add constant speed up support to veJoeStaking

0xnivek:
This PR adds support for speed up to `veJoeStaking` as well as appropriate tests.

Some important things to note:
- User will receive speed up benefits if they are depositing for first time OR they deposit `speedUpThreshold` percentage of their current amount of staked JOE
- The speed up benefit is **on top** of the user's existing base veJoe accumulation
- The speed up generation rate is **constant and cannot be changed**

### DIFF
--- a/contracts/VeJoeStaking.sol
+++ b/contracts/VeJoeStaking.sol
@@ -31,7 +31,7 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
         uint256 lastClaimTimestamp;
         uint256 speedUpEndTimestamp;
         /**
-         * @notice We do some fancy math here. Basically, any point in time, the amount of JOEs
+         * @notice We do some fancy math here. Basically, any point in time, the amount of veJOE
          * entitled to a user but is pending to be distributed is:
          *
          *   pendingReward = pendingBaseReward + pendingSpeedUpReward
@@ -109,18 +109,12 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
         VeJoeToken _veJoe,
         uint256 _veJoePerSharePerSec,
         uint256 _speedUpVeJoePerSharePerSec,
-        uint256 _maxCap,
         uint256 _speedUpThreshold,
-        uint256 _speedUpDuration
+        uint256 _speedUpDuration,
+        uint256 _maxCap,
     ) public initializer {
         require(address(_joe) != address(0), "VeJoeStaking: unexpected zero address for _joe");
         require(address(_veJoe) != address(0), "VeJoeStaking: unexpected zero address for _veJoe");
-
-        upperLimitMaxCap = 100000;
-        require(
-            _maxCap != 0 && _maxCap <= upperLimitMaxCap,
-            "VeJoeStaking: expected new _maxCap to be non-zero and <= 100000"
-        );
 
         require(
             _speedUpThreshold != 0 && _speedUpThreshold <= 100,
@@ -131,6 +125,12 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
         require(
             _speedUpDuration <= upperLimitSpeedUpDuration,
             "VeJoeStaking: expected _speedUpDuration to be <= 365 days"
+        );
+
+        upperLimitMaxCap = 100000;
+        require(
+            _maxCap != 0 && _maxCap <= upperLimitMaxCap,
+            "VeJoeStaking: expected new _maxCap to be non-zero and <= 100000"
         );
 
         __Ownable_init();
--- a/contracts/VeJoeStaking.sol
+++ b/contracts/VeJoeStaking.sol
@@ -63,6 +63,9 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
     /// @notice veJOE per sec per JOE staked, scaled to `VEJOE_PER_SHARE_PER_SEC_PRECISION`
     uint256 public veJoePerSharePerSec;
 
+    /// @notice speed up veJOE per sec per JOE staked, scaled to `VEJOE_PER_SHARE_PER_SEC_PRECISION`
+    uint256 public speedUpVeJoePerSharePerSec;
+
     /// @notice Precision of `veJoePerSharePerSec`
     uint256 public VEJOE_PER_SHARE_PER_SEC_PRECISION;
 
@@ -99,6 +102,7 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
         IERC20Upgradeable _joe,
         VeJoeToken _veJoe,
         uint256 _veJoePerSharePerSec,
+        uint256 _speedUpVeJoePerSharePerSec,
         uint256 _maxCap,
         uint256 _speedUpThreshold,
         uint256 _speedUpDuration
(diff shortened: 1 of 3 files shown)


contracts/VeJoeStaking.sol
```diff
@@ -46,35 +66,70 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
     uint256 public lastRewardTimestamp;
 
-    /// @notice veJOE per sec per JOE staked, scaled to `VEJOE_PER_SEC_PRECISION`
-    uint256 public veJoePerSec;
+    /// @notice veJOE per sec per JOE staked, scaled to `VEJOE_PER_SHARE_PER_SEC_PRECISION`
+    uint256 public veJoePerSharePerSec;
 
-    /// @notice Precision of `veJoePerSec`
-    uint256 public VEJOE_PER_SEC_PRECISION;
+    /// @notice speed up veJOE per sec per JOE staked, scaled to `VEJOE_PER_SHARE_PER_SEC_PRECISION`
```
cryptofish7, line 72: `speed`->`Speed`

contracts/VeJoeStaking.sol
```diff
@@ -119,12 +199,38 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
 
-        UserInfo storage userInfo = userInfos[msg.sender];
+            uint256 userStakedJoe = userInfo.balance;
+
+            uint256 userVeJoeBalance = veJoe.balanceOf(msg.sender);
+            uint256 userMaxVeJoeCap = userStakedJoe.mul(maxCap);
+
+            // If the user is currently at their max veJOE cap, we need to update
+            // their `lastClaimTimestamp` to now to prevent passive veJOE accrual
+            // after user hit their max cap.
```
cryptofish7, line 216: `hit`->`hits`

contracts/VeJoeStaking.sol
```diff
@@ -178,19 +286,33 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
             );
         }
-
-        uint256 pendingVeJoe = _accVeJoePerShare.mul(user.balance).div(ACC_VEJOE_PER_SHARE_PRECISION).sub(
+        uint256 pendingBaseVeJoe = _accVeJoePerShare.mul(user.balance).div(ACC_VEJOE_PER_SHARE_PRECISION).sub(
             user.rewardDebt
         );
 
+        // Calculate amount of pending speed up veJOE
+        uint256 pendingSpeedUpVeJoe = 0;
```
cryptofish7, line 304: Change to `uint256 pendingSpeedUpVeJoe;` for tiny gas savings

cryptofish7 (changes requested):

contracts/VeJoeStaking.sol
```diff
@@ -178,19 +286,33 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
+        if (user.speedUpEndTimestamp != 0) {
```
LouisMeMyself, line 299: There is an issue, if we change the `speedUpDuration`, current user that have a speedup won't have any benefit for it.
If we want to account for this, I suggest we add a `startSpeedUpTimestamp`, and before anything we test that `endSpeedUpTimestamp - startSpeedUpTimestamp == speedUpDuration`, if it's not the case, we change `endSpeedUpTimestamp` to account for the new changes.
Issue is that if we reduce the duration, then some users may have harvested before and thus would have had some `speedUp` rewards that they won't have if they'd have harvested only after the changes

LouisMeMyself (changes requested):
2 small issues

contracts/VeJoeStaking.sol
```diff
@@ -119,12 +199,38 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
+
+            // User is eligible for speed up benefits if `_amount` is at least
+            // `speedUpThreshold / 100 * userStakedJoe`
+            if (_amount.mul(100) >= speedUpThreshold.mul(userStakedJoe)) {
+                userInfo.speedUpEndTimestamp = block.timestamp.add(speedUpDuration);
+            }
+        } else {
+            // If the user's `lastClaimTimestamp` is 0, i.e. if this is the user's first time staking,
+            // then they will receive speed up benefits.
+            if (userInfo.lastClaimTimestamp == 0) {
```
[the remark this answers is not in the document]
0xnivek: Fixed!

contracts/VeJoeStaking.sol
0xnivek: As per conversation in discord, aligned on making `speedUpDuration` constant

cryptofish7 (approved):
LGTM

contracts/VeJoeStaking.sol
```diff
@@ -46,35 +66,62 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
     ) public initializer {
         require(address(_joe) != address(0), "VeJoeStaking: unexpected zero address for _joe");
         require(address(_veJoe) != address(0), "VeJoeStaking: unexpected zero address for _veJoe");
 
+        require(
+            _speedUpThreshold != 0 && _speedUpThreshold <= 100,
+            "VeJoeStaking: expected _speedUpThreshold to be > 0 and <= 100"
+        );
+
+        require(_speedUpDuration <= 365 days, "VeJoeStaking: expected _speedUpDuration to be <= 365 days");
```
jummy123, line 123: Is it intended for 0 to be a valid value here? I assume if we wanted to disable sped up generation we set to 0.

contracts/VeJoeStaking.sol
```diff
@@ -104,12 +154,23 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
-    /// @notice Set veJoePerSec
-    /// @param _veJoePerSec The new veJoePerSec
-    function setVeJoePerSec(uint256 _veJoePerSec) external onlyOwner {
+    /// @notice Set veJoePerSharePerSec
+    /// @param _veJoePerSharePerSec The new veJoePerSharePerSec
+    function setVeJoePerSharePerSec(uint256 _veJoePerSharePerSec) external onlyOwner {
         updateRewardVars();
-        veJoePerSec = _veJoePerSec;
-        emit UpdateVeJoePerSec(msg.sender, _veJoePerSec);
+        veJoePerSharePerSec = _veJoePerSharePerSec;
```
jummy123, line 161: Probably good to have some upper bound on this (it can be high), it won't permanently break the contract if we set it stupidly high but could cause overflow on line 265, 
`secondsElapsed.mul(veJoePerSharePerSec).mul(ACC_VEJOE_PER_SHARE_PRECISION)` could cause phantom overflow.

jummy123: See phantom overflow in this great series of articles [math in solidity](https://medium.com/coinmonks/math-in-solidity-part-2-overflow-3cd7283714b4)

contracts/VeJoeStaking.sol
```diff
@@ -119,12 +180,28 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
+
+            uint256 userStakedJoe = userInfo.balance;
+
+            // User is eligible for speed up benefits if `_amount` is at least
+            // `speedUpThreshold / 100 * userStakedJoe`
+            if (_amount.mul(100) >= speedUpThreshold.mul(userStakedJoe)) {
+                userInfo.speedUpEndTimestamp = block.timestamp.add(speedUpDuration);
+            }
+        } else {
+            userInfo.speedUpEndTimestamp = block.timestamp.add(speedUpDuration);
```
jummy123, line 202: A comment here could be useful saying if a user is making a deposit with no current balance they automatically get the speed up.

contracts/VeJoeStaking.sol
0xnivek: yeah I don't feel too strongly about this. I wanted to leave it open as an option in case we did want to set it to 0 in the future

contracts/VeJoeStaking.sol
0xnivek: Added a `upperLimitVeJoePerSharePerSec = 1e36`. thanks for the article, was a good read!

PR #44, opened 2022-02-21
Some minor cleanup of StableJoeStaking

0xnivek:


contracts/StableJoeStaking.sol
```diff
@@ -131,16 +135,19 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
             IERC20Upgradeable _token = rewardTokens[i];
             updateReward(_token);
 
+            uint256 _previousRewardDebt = user.rewardDebt[_token];
+            user.rewardDebt[_token] = _newAmount.mul(accRewardPerShare[_token]).div(ACC_REWARD_PER_SHARE_PRECISION);
```
0xnivek, line 139: this change is to make sure we follow checks-effects-interaction pattern before we perform the `safeTokenTransfer` in line 147

contracts/StableJoeStaking.sol
```diff
@@ -299,16 +309,16 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
         uint256 _currRewardBalance = _token.balanceOf(address(this));
         uint256 _rewardBalance = _token == joe ? _currRewardBalance.sub(internalJoeBalance) : _currRewardBalance;
 
-        uint256 _totalJoe = internalJoeBalance;
```
LouisMeMyself, line 302: This line may seem useless, but in fact it's some gas optimization, when you use a var more than once it costs less gas to declare a new var copying that value. As we're using `updateReward` inside a loop, we need to take care of gas, so I'd be in favor of keeping that line and using in the next 2 calls

contracts/StableJoeStaking.sol
```diff
@@ -224,17 +231,19 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
     function pendingReward(address _user, IERC20Upgradeable _token) external view returns (uint256) {
         require(isRewardToken[_token], "StableJoeStaking: wrong reward token");
         UserInfo storage user = userInfo[_user];
-        uint256 _totalJoe = internalJoeBalance;
```
LouisMeMyself, line 227: Here it's not important as it's a view function, external it will be free either way

LouisMeMyself (changes requested):

contracts/StableJoeStaking.sol
0xnivek: ahh I see, that makes sense! fixed

cryptofish7 (approved):

PR #47, opened 2022-02-22
WIP: Vejoe deploy

jummy123:
@cryptofish7 some questions on how MVC3 is deployed.
1) Do we need to create the pool for rewards on MVC2 before the deploy of boostedMC?
2) It seems the 0x66F address is the owner of MCV2 and I don't really want to put the `impersonate_account` stuff in the deploy files. Should we leave creating the pool on MCV2 and calling the `init` function on BMCJ to manually after the deploy. (only issue I see with this is the PID argument to BMCJ).


deploy/BoostedMasterChefJoe.js
```diff
@@ -0,0 +1,58 @@
...
+    joeAddress = "0xce347E069B68C53A9ED5e7DA5952529cAF8ACCd4";
+    masterChefV2Address = "0x1F51b7697A1919cF301845c93D4843FD620ad7Cc";
+  } else if (chainId == 43114 || chainId == 31337) {
+    // avalanche mainnet or hardhat network addresses
+    joeAddress = "0x6e84a6216eA6dACC71eE8E6b0a5B7322EEbC0fDd";
+    masterChefV2Address = "0xC98C3C547DDbcc0029F38E0383C645C202aD663d";
+  }
+
+  const MCV2 = await ethers.getContract("MasterChefJoeV2", masterChefV2Address);
+  await (await MCV2.add(100, dummyToken.address, false)).wait();
```
jummy123, line 30: Currently owner of this contract is 0x66 address so this step needs to be completed manually.

PR #65, opened 2022-03-02
[VeJoeStaking] Update maxCap to be maxCapPct (scaled to 100)

0xnivek:
Currently, `VeJoeStaking` has a notion of `maxCap` which is a whole number that represents the max veJOE a user can have, e.g. if `maxCap = 100`, `maxVeJoeBalance = user.balance * 100`. However, I was looking at the `Staking V2 Contract Specs` and noticed max cap was being referred to as:

 `MAX_CAP_PCT = Max limit of veJOE as percentage points of JOE balance`

This means that we should allow setting a user's max `veJOE` to be `1.5X` their current staked `JOE`, which you wouldn't be able to currently do. This PR updates `VeJoeStaking` to handle this

### DIFF
--- a/contracts/VeJoeStaking.sol
+++ b/contracts/VeJoeStaking.sol
@@ -157,7 +157,7 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
 
     /// @notice Set maxCapPct
     /// @param _maxCapPct The new maxCapPct
-    function setMaxCap(uint256 _maxCapPct) external onlyOwner {
+    function setMaxCapPct(uint256 _maxCapPct) external onlyOwner {
         require(_maxCapPct > maxCapPct, "VeJoeStaking: expected new _maxCapPct to be greater than existing maxCapPct");
         require(
             _maxCapPct != 0 && _maxCapPct <= upperLimitMaxCapPct,
--- a/contracts/VeJoeStaking.sol
+++ b/contracts/VeJoeStaking.sol
@@ -115,6 +115,8 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
         uint256 _speedUpDuration,
         uint256 _maxCapPct
     ) public initializer {
+        __Ownable_init();
+
         require(address(_joe) != address(0), "VeJoeStaking: unexpected zero address for _joe");
         require(address(_veJoe) != address(0), "VeJoeStaking: unexpected zero address for _veJoe");
 
@@ -140,8 +142,6 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
             _maxCapPct != 0 && _maxCapPct <= upperLimitMaxCapPct,
             "VeJoeStaking: expected _maxCapPct to be non-zero and <= 10000000"
         );
-
-        __Ownable_init();
 
         maxCapPct = _maxCapPct;
         speedUpThreshold = _speedUpThreshold;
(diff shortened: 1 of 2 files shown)


contracts/VeJoeStaking.sol
```diff
@@ -305,7 +305,7 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
 
         // Get the user's current veJOE balance and maximum veJOE they can hold
         uint256 userVeJoeBalance = veJoe.balanceOf(_user);
-        uint256 userMaxVeJoeCap = user.balance.mul(maxCap);
+        uint256 userMaxVeJoeCap = user.balance.mul(maxCapPct).div(100);
```
0xnivek, line 308: note: I realized doing the calculation like this means that if `maxCapPct < 100`, e.g. `50`, the user would need to have at least `2` JOE staked but I think this is fine given that:

1. JOE and veJOE are 18 decimals. 
2. In practice, I don't think we will ever run into this case since I assume we are going to set `maxCapPct` to an initial value `>= 100` which means we won't be able to ever set it to be smaller than that
3. Even if we did want to handle this edge case, i think it would be difficult to handle it for logic in lines 313-314

curious if everyone agrees

contracts/VeJoeStaking.sol
```diff
@@ -135,15 +135,15 @@ contract VeJoeStaking is Initializable, OwnableUpgradeable {
...
-        upperLimitMaxCap = 100000;
+        upperLimitMaxCapPct = 10000000;
         require(
-            _maxCap != 0 && _maxCap <= upperLimitMaxCap,
-            "VeJoeStaking: expected _maxCap to be non-zero and <= 100000"
+            _maxCapPct != 0 && _maxCapPct <= upperLimitMaxCapPct,
+            "VeJoeStaking: expected _maxCapPct to be non-zero and <= 10000000"
         );
 
         __Ownable_init();
```
LouisMeMyself, line 144: Inits should always be the first line of code of initializer.
Cause during the constructor we could execute code and `owner` wouldn't be set yet

contracts/VeJoeStaking.sol
LouisMeMyself: They wouldn't need 2 `JOE`, they would need 2e-18 `JOE` which is definitely fine.
This could be fixed with some basic math tho, but not needed.

Just for the record, something like that:
```
        uint256 userMaxVeJoeCap = user.balance.mul(maxCapPct); // this is 100 time bigger than expected
        if (userVeJoeBalance.mul(100) >= userMaxVeJoeCap) {
            // User already holds maximum amount of veJOE so there is no pending veJOE
            return 0;
        } else if (userVeJoeBalance.add(pendingVeJoe).mul(100) > userMaxVeJoeCap) {
            return userMaxVeJoeCap.sub(userVeJoeBalance.mul(100)).div(100);
        } else {
            return pendingVeJoe;
        }
```

LouisMeMyself (approved):

jummy123:
The changes look good to me but as this contract is currently being audited should we hold off on the merge? cc @cryptofish7

0xnivek:
> The changes look good to me but as this contract is currently being audited should we hold off on the merge? cc @cryptofish7

makes sense, will wait to land this

cryptofish7 (approved):

PR #70, opened 2022-03-10
Set speedUpDuration to 15 days in VeJoeStaking deploy

0xnivek:
Noticed its currently set to 15 seconds rather than 15 days


jummy123 (approved):
LGTM, once merged I will deploy today. Do you think we need to add a function to update this variable? `setSpeedUpDuration(uint256 seconds)`?

0xnivek:
> LGTM, once merged I will deploy today. Do you think we need to add a function to update this variable? `setSpeedUpDuration(uint256 seconds)`?

Supporting dynamic `speedUpDuration` would require some additional logic changes and from speaking w/ murloc it seems like we're fine keeping it as static value

PR #74, opened 2022-03-11
Add UpdateBoostedMasterChefJoe event to VeJoeToken

0xnivek:


jummy123 (approved):

cryptofish7 (approved):

contracts/VeJoeToken.sol
```diff
@@ -36,6 +38,8 @@ contract VeJoeToken is VeERC20("VeJoeToken", "veJOE"), Ownable {
     function setBoostedMasterChefJoe(address _boostedMasterChef) external onlyOwner {
         // We allow 0 address here if we want to disable the callback operations
         boostedMasterChef = IBoostedMasterChefJoe(_boostedMasterChef);
+
+        emit UpdateBoostedMasterChefJoe(msg.sender, _boostedMasterChef);
```
cryptofish7, line 42: Use `_msgSender()`

cryptofish7 (changes requested):

contracts/VeJoeToken.sol
jummy123: This contract doesn't inherit from `Context` so unless we change that we can't use `_msgSender` here.

LouisMeMyself: It does inherit from `Context` as `Ownable` inherits it.

LouisMeMyself (approved):

PR #110, opened 2022-04-01
gas opt

LouisMeMyself:


cryptofish7 (changes requested):
Something very weird in this PR. Check the constructor function; it has a nested modifier and constructor function. Also weird that it didn't pick this up in the diff

cryptofish7 (approved):

PR #117, opened 2022-09-02
No deposit fee

cryptofish7:
In this PR I add feature to sJOE that exempts EOAs holding a Smol Joe from paying the deposit fee.

### DIFF
--- a/contracts/StableJoeStaking.sol
+++ b/contracts/StableJoeStaking.sol
@@ -107,15 +107,18 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
         IERC20Upgradeable _rewardToken,
         IERC20Upgradeable _joe,
         address _feeCollector,
-        uint256 _depositFeePercent
+        uint256 _depositFeePercent,
+        IERC721Upgradeable _smolJoes
     ) external initializer {
         __Ownable_init();
         require(address(_rewardToken) != address(0), "StableJoeStaking: reward token can't be address(0)");
         require(address(_joe) != address(0), "StableJoeStaking: joe can't be address(0)");
+        require(address(_smolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)");
         require(_feeCollector != address(0), "StableJoeStaking: fee collector can't be address(0)");
         require(_depositFeePercent <= 5e17, "StableJoeStaking: max deposit fee can't be greater than 50%");
 
         joe = _joe;
+        smolJoes = _smolJoes;
         depositFeePercent = _depositFeePercent;
         feeCollector = _feeCollector;
 
@@ -134,7 +137,7 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
 
         uint256 _fee;
         // Only EOAs holding Smol Joes are exempt from paying the deposit fee
-        if (smolJoes.balanceOf(_msgSender()) == 0 || _msgSender() != tx.origin) {
+        if ((address(smolJoes) != 0 && smolJoes.balanceOf(_msgSender()) == 0) || _msgSender() != tx.origin) {
             _fee = _amount.mul(depositFeePercent).div(DEPOSIT_FEE_PERCENT_PRECISION);
         }
         uint256 _amountMinusFee = _amount.sub(_fee);
@@ -237,11 +240,12 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
 
     /**
      * @notice Initialize the Smol Joes address
+     * @dev This function was added to be able to set Smol Joes during an upgrade
+     * as the contract was already initialized
      * @param _smolJoes The Smol Joes contract address
      */
-    function initializeSmolJoes(IERC721Upgradeable _smolJoes) external onlyOwner {
-        require(address(_smolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)");
-        require(address(smolJoes) == address(0), "StableJoeStaking: smol joes already initialized");
+    function setSmolJoes(IERC721Upgradeable _smolJoes) external onlyOwner {
+        require(address(_smolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)3");
         smolJoes = _smolJoes;
         emit SmolJoesInitialized(_smolJoes, smolJoes);
     }
--- a/contracts/StableJoeStaking.sol
+++ b/contracts/StableJoeStaking.sol
@@ -92,7 +92,7 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
     event RewardTokenRemoved(address token);
 
     /// @notice Emitted when Smol Joes address is changed
-    event SmolJoesChanged(IERC721Upgradeable newSmolJoes, IERC721Upgradeable oldSmolJoes);
+    event SmolJoesInitialized(IERC721Upgradeable newSmolJoes, IERC721Upgradeable oldSmolJoes);
 
     /**
      * @notice Initialize a new StableJoeStaking contract
@@ -102,24 +102,20 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
      * @param _joe The address of the JOE token
      * @param _feeCollector The address where deposit fees will be sent
      * @param _depositFeePercent The deposit fee percent, scalled to 1e18, e.g. 3% is 3e16
-     * @param _smolJoes The address of ERC721 Smol Joes NFT
      */
     function initialize(
         IERC20Upgradeable _rewardToken,
         IERC20Upgradeable _joe,
         address _feeCollector,
-        uint256 _depositFeePercent,
-        IERC721Upgradeable _smolJoes
+        uint256 _depositFeePercent
     ) external initializer {
         __Ownable_init();
         require(address(_rewardToken) != address(0), "StableJoeStaking: reward token can't be address(0)");
         require(address(_joe) != address(0), "StableJoeStaking: joe can't be address(0)");
-        require(address(_smolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)");
         require(_feeCollector != address(0), "StableJoeStaking: fee collector can't be address(0)");
         require(_depositFeePercent <= 5e17, "StableJoeStaking: max deposit fee can't be greater than 50%");
 
         joe = _joe;
-        smolJoes = _smolJoes;
         depositFeePercent = _depositFeePercent;
         feeCollector = _feeCollector;
 
@@ -240,14 +236,14 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
     }
 
     /**
-     * @notice Set the Smol Joes address
-     * @param _newSmolJoes The new Smol Joes contract address
+     * @notice Initialize the Smol Joes address
+     * @param _smolJoes The Smol Joes contract address
      */
-    function setSmolJoes(IERC721Upgradeable _newSmolJoes) external onlyOwner {
-        require(address(_newSmolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)");
-        IERC721Upgradeable oldSmolJoes = smolJoes;
-        smolJoes = _newSmolJoes;
-        emit SmolJoesChanged(_newSmolJoes, oldSmolJoes);
+    function initializeSmolJoes(IERC721Upgradeable _smolJoes) external onlyOwner {
+        require(address(_smolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)");
+        require(address(smolJoes) == address(0), "StableJoeStaking: smol joes already initialized");
+        smolJoes = _smolJoes;
+        emit SmolJoesInitialized(_smolJoes, smolJoes);
     }
 
     /**
--- a/contracts/StableJoeStaking.sol
+++ b/contracts/StableJoeStaking.sol
@@ -68,7 +68,7 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
     mapping(address => UserInfo) private userInfo;
 
     /// @dev Smol Joes contract address
-    IERC721Upgradeable smolJoes;
+    IERC721Upgradeable public smolJoes;
 
     /// @notice Emitted when a user deposits JOE
     event Deposit(address indexed user, uint256 amount, uint256 fee);
--- a/contracts/StableJoeStaking.sol
+++ b/contracts/StableJoeStaking.sol
@@ -137,7 +137,7 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
 
         uint256 _fee;
         // Only EOAs holding Smol Joes are exempt from paying the deposit fee
-        if ((address(smolJoes) != 0 && smolJoes.balanceOf(_msgSender()) == 0) || _msgSender() != tx.origin) {
+        if (address(smolJoes) != 0 && (smolJoes.balanceOf(_msgSender()) == 0 || _msgSender() != tx.origin)) {
             _fee = _amount.mul(depositFeePercent).div(DEPOSIT_FEE_PERCENT_PRECISION);
         }
         uint256 _amountMinusFee = _amount.sub(_fee);
@@ -245,7 +245,7 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
      * @param _smolJoes The Smol Joes contract address
      */
     function setSmolJoes(IERC721Upgradeable _smolJoes) external onlyOwner {
-        require(address(_smolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)3");
+        require(address(_smolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)");
         smolJoes = _smolJoes;
         emit SmolJoesInitialized(_smolJoes, smolJoes);
     }
--- a/contracts/StableJoeStaking.sol
+++ b/contracts/StableJoeStaking.sol
@@ -91,6 +91,9 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
     /// @notice Emitted when owner removes a token from the reward tokens list
     event RewardTokenRemoved(address token);
 
+    /// @notice Emitted when Smol Joes address is changed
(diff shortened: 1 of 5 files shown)


deploy/StableJoeStaking.js
```diff
@@ -12,6 +17,7 @@ module.exports = async function ({ getNamedAccounts, deployments }) {
     rewardToken = "0x9Ad6C38BE94206cA50bb0d90783181662f0Cfa10";
     feeCollector = "0x2fbB61a10B96254900C03F1644E9e1d2f5E76DD2";
     depositFeePercent = 0;
+    smolJoes = "0x0000000000000000000000000000000000000000"; // placeholder for now
```
0x0Louis, line 20: This will revert on initialize, use any ERC20 contract as they also define `balanceOf` function or use a proper mock NFT contract

contracts/StableJoeStaking.sol
```diff
@@ -125,7 +133,11 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
     function deposit(uint256 _amount) external {
         UserInfo storage user = userInfo[_msgSender()];
 
-        uint256 _fee = _amount.mul(depositFeePercent).div(DEPOSIT_FEE_PERCENT_PRECISION);
+        uint256 _fee;
+        // Only EOAs holding Smol Joes are exempt from paying the deposit fee
+        if (smolJoes.balanceOf(_msgSender()) == 0 || _msgSender() != tx.origin) {
```
0x0Louis, line 138: This will revert on `smolJoes` being address 0, (this will be the case during upgrade)

contracts/StableJoeStaking.sol
```diff
@@ -95,20 +99,24 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
...
      * @param _depositFeePercent The deposit fee percent, scalled to 1e18, e.g. 3% is 3e16
+     * @param _smolJoes The address of ERC721 Smol Joes NFT
      */
     function initialize(
         IERC20Upgradeable _rewardToken,
         IERC20Upgradeable _joe,
         address _feeCollector,
-        uint256 _depositFeePercent
+        uint256 _depositFeePercent,
+        IERC721Upgradeable _smolJoes
```
0x0Louis, line 109: Add a setter for smolJoe's address as we can't use initialize during an upgrade as the contract has already been initialized.

deploy/StableJoeStaking.js
```diff
@@ -32,7 +39,13 @@ module.exports = async function ({ getNamedAccounts, deployments }) {
...
+            args: [
+              rewardToken,
+              joeAddress,
+              feeCollector,
+              depositFeePercent,
+              smolJoes,
+            ],
           },
         },
       },
```
0x0Louis, line 51: For safety reasons, don't forget to initialize the implementation even if for transparent proxy this is useless.

0x0Louis (changes requested):
Small nits.
Please add a non zero check at L168 and L169 so it stops spamming the multi sig with transfer(0). (it will also decrease gas cost)

cryptofish7:
> Small nits. Please add a non zero check at L168 and L169 so it stops spamming the multi sig with transfer(0). (it will also decrease gas cost)

What do you mean by this?

0x0Louis:
I mean to add a:
```solidity
if (amount > 0) token.safeTransferFrom(msg.sender, to, amount);
```
at L168 and 169 for the 2 different transfers

0x0Louis (approved):

contracts/StableJoeStaking.sol
```diff
@@ -102,24 +102,20 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
...
     ) external initializer {
         __Ownable_init();
         require(address(_rewardToken) != address(0), "StableJoeStaking: reward token can't be address(0)");
         require(address(_joe) != address(0), "StableJoeStaking: joe can't be address(0)");
-        require(address(_smolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)");
         require(_feeCollector != address(0), "StableJoeStaking: fee collector can't be address(0)");
         require(_depositFeePercent <= 5e17, "StableJoeStaking: max deposit fee can't be greater than 50%");
 
         joe = _joe;
-        smolJoes = _smolJoes;
```
0x0Louis: `initialize` should still set smolJoes' address.
One easy way could be to make `initializeSmolJoes` public and use it here as well

contracts/StableJoeStaking.sol
```diff
@@ -240,14 +236,14 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
     }
 
     /**
-     * @notice Set the Smol Joes address
-     * @param _newSmolJoes The new Smol Joes contract address
+     * @notice Initialize the Smol Joes address
+     * @param _smolJoes The Smol Joes contract address
```
0x0Louis, line 240: Add a comment explaining that this function was added to be able to set smolJoe during an upgrade, as the contract was already initialized

deploy/StableJoeStaking.js
```diff
@@ -38,14 +38,8 @@ module.exports = async function ({ getNamedAccounts, deployments }) {
...
-            methodName: "initialize",
-            args: [
-              rewardToken,
-              joeAddress,
-              feeCollector,
-              depositFeePercent,
-              smolJoes,
-            ],
+            methodName: "initializeSmolJoes",
+            args: [smolJoes],
```
0x0Louis, line 42: This should still call `initialize`, the `initializeSmolJoes` is only for an upgrade, and we need to call it by the multisig

0x0Louis (changes requested):

deploy/StableJoeStaking.js
```diff
@@ -12,33 +17,45 @@ module.exports = async function ({ getNamedAccounts, deployments }) {
...
+    const sJoe = await deploy("StableJoeStaking", {
       from: deployer,
       proxy: {
         owner: proxyOwner,
         proxyContract: "OpenZeppelinTransparentProxy",
         execute: {
           init: {
-            methodName: "initialize",
-            args: [rewardToken, joeAddress, feeCollector, depositFeePercent],
+            methodName: "initializeSmolJoes",
```
jummy123, line 41: You shouldn't change this block (`init`) as it won't get called on upgrade anyway (plus it could have unintended consequences for the unit tests if they use these deployment scripts for fixture generation). If you want to use hardhat-deploy to set the address you can add an `onUpgrade` block

```
    init: {...unchanged...},
    onUpgrade: {
        methodName: "initializeSmolJoes",
        args: [smolJoes],
    }
```
To do this however you will need to make the `initializeSmolJoes` function idempotent (not reverting if `smolJoes` variable is non 0) as it will be called every time the function gets upgraded.

contracts/StableJoeStaking.sol
jummy123: I believe is talking about the case between when the contract is upgraded and the `initializeSmalJoes` function is called (which may be a very small amount of time). You could do this in a backwards compatible way by changing this to 
`if (address(smolJoes) != 0 && smolJoes.balanceOf(_msgSender()) == 0 || _msgSender() != tx.origin)` this works due to the [short circuiting](https://docs.soliditylang.org/en/latest/types.html#booleans) behaviour of `&&` and `||`

contracts/StableJoeStaking.sol
```diff
@@ -224,6 +235,17 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
         emit DepositFeeChanged(_depositFeePercent, oldFee);
     }
 
+    /**
+     * @notice Initialize the Smol Joes address
+     * @param _smolJoes The Smol Joes contract address
+     */
+    function initializeSmolJoes(IERC721Upgradeable _smolJoes) external onlyOwner {
```
jummy123, line 242: Is there a reason you only allow setting this value once? From a purely pragmatic point of view if there is some mistake in the `smolJoes` address the contract will be permanently broken and you won't be able to call it again with the correct address.

deploy/StableJoeStaking.js
0x0Louis: That's very cool, I didn't know. Thanks for sharing that!
However, if we remove that check, then anyone can set it to any address. The new `Initializable` (https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/proxy/utils/Initializable.sol) from OZ makes a lot of sense in that case.

contracts/StableJoeStaking.sol
0x0Louis: That's not wrong, but as it's a proxy nothing is permanent in any case.

deploy/StableJoeStaking.js
cryptofish7: I tried to use `onUpgrade` but it would revert without any reason when I ran the upgrade locally. On the contrary, when I didn't use `onUpgrade` it would upgrade fine but obviously I had to call `initializeSmolJoes` manually:

Hey, I have a question about proxy upgrades. I have a proxy contract `Test` that I want to upgrade. On the upgrade, I want it to call a function `intitializeAgain()`.

In my deploy script I have:

```
    await deploy("Test", {
      from: deployer,
      proxy: {
        owner: deployer,
        proxyContract: "OpenZeppelinTransparentProxy",
        execute: {
          init: {
            methodName: "initialize",
            args: [
              param1,
              param2,
            ],
          },
          onUpgrade: {
            methodName: "initializeAgain",
            args: [param3],
          },
        },
      },
      log: true,
    });
```

When I run this locally and try to upgrade the implementation after making a modification to the `Test` contract, it reverts without any reason:

```
 Compilation finished successfully
Creating Typechain artifacts in directory types for target ethers-v5
Successfully generated Typechain artifacts!
deployer is:  0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
proxyOwner is:  0x2fbB61a10B96254900C03F1644E9e1d2f5E76DD2
reusing "DefaultProxyAdmin" at 0x9fE46736679d2D9a65F0992F2272dE9f3c7fa6e0
deploying "StableJoeStaking_Implementation" (tx: 0x44ad6abb7bc903b949f19176000973c787c83e1b16f92ebbca42e6a9cec86934)...: deployed at 0x0165878A594ca255338adfa4d48449f69242Eb8F with 3619599 gas
An unexpected error occurred:

Error: ERROR processing /Users/cryptofish/workspace/joe-core/deploy/StableJoeStaking.js:
ProviderError: Error: Transaction reverted without a reason string
```

The deploy code looks like this:

```
    const sJoe = await deploy("StableJoeStaking", {
      from: deployer,
      proxy: {
        owner: deployer, // change from proxyOwner so there are no ownership issues on upgrading
        proxyContract: "OpenZeppelinTransparentProxy",
        execute: {
          init: {
            methodName: "initialize",
            args: [
              rewardToken,
              joeAddress,
              feeCollector,
              depositFeePercent,
              smolJoes,
            ],
          },
          onUpgrade: {
            methodName: "initializeSmolJoes",
            args: [smolJoes],
          },
        },
      },
      log: true,
    });

```

jummy123: So it looks like this is failing because of the `onlyOwner` modifier on `initializeSmolJoes`. This is because hardhat-deploy will use `upgradeToAndCall` function on the `ProxyAdmin` contract which performs the upgrade and then calls `initializeSmolJoes`. As the `ProxyAdmin` address isn't the owner of the `StableJoeStaking` contract this fails. I confirmed this by calling `transferOwnership` with the `ProxyAdmin` address on `StableJoeStaking` and then performing the upgrade, in this case the `initializeSmolJoes` function didn't revert as `msg.sender` was the owner (The proxy admin address). This makes sense as the `upgradeToAndCall` function should be atomic (the upgrade and function call should succeed or the tx should rollback). 

AFAICT there is two options here:
1. Change the ownership of `StableJoeStaking` to the `ProxyAdmin` contract.
2. Not use atomic upgrade and call the `initializeSmolJoes` function after the contract has been upgraded, this option is probably fine as the smolJoes address **isn't** needed for the contract to function.

cryptofish7: Thanks for spotting that. Regarding the two options - option 1 is a pain since several other functions are `onlyOwner` so you'll need to transfer ownership of `StableJoeStaking` back to the multisig. 

Can't put `initializeSmolJoes` under the `initializer` modifier either. If only we could use `reinitializer` in the latest OZ `Initializable`, but it's v8 so not compatible. 

Anyways, gone with option 2 here and turned `initializeSmolJoes` to `setSmolJoes`.

0x0Louis: Option 1 would be a big mistake: Transparent Proxies don't allow the owner of the proxy to call any function of the implementation. So if the ownership is transferred, you couldn't call any `onlyOwner` function, nor transfer the ownership again.

contracts/StableJoeStaking.sol
```diff
@@ -134,7 +137,7 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
 
         uint256 _fee;
         // Only EOAs holding Smol Joes are exempt from paying the deposit fee
-        if (smolJoes.balanceOf(_msgSender()) == 0 || _msgSender() != tx.origin) {
+        if ((address(smolJoes) != 0 && smolJoes.balanceOf(_msgSender()) == 0) || _msgSender() != tx.origin) {
```
0x0Louis, line 140: Wrong parenthesis, if smolJoe is not set EOAs wouldn't pay any fees, change to:
`if (address(smolJoes) != 0 && (smolJoes.balanceOf(_msgSender()) == 0 || _msgSender() != tx.origin)) {`

contracts/StableJoeStaking.sol
```diff
@@ -237,11 +240,12 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
...
      * @notice Initialize the Smol Joes address
+     * @dev This function was added to be able to set Smol Joes during an upgrade
+     * as the contract was already initialized
      * @param _smolJoes The Smol Joes contract address
      */
-    function initializeSmolJoes(IERC721Upgradeable _smolJoes) external onlyOwner {
-        require(address(_smolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)");
-        require(address(smolJoes) == address(0), "StableJoeStaking: smol joes already initialized");
+    function setSmolJoes(IERC721Upgradeable _smolJoes) external onlyOwner {
+        require(address(_smolJoes) != address(0), "StableJoeStaking: smol joes can't be address(0)3");
```
0x0Louis, line 248: Typo address(0)**3**

deploy/StableJoeStaking.js
```diff
@@ -38,22 +38,34 @@ module.exports = async function ({ getNamedAccounts, deployments }) {
...
         rewardToken,
         joeAddress,
         feeCollector,
-        depositFeePercent
+        depositFeePercent,
+        smolJoes
       );
+      console.log("Setting Smol Joes...");
+      const sJoeProxy = await ethers.getContract("StableJoeStaking");
+      await sJoeProxy.setSmolJoes(smolJoes);
```
0x0Louis, line 68: If it's newly deployed then this is not necessary, as it would set it with initialize

0x0Louis (changes requested):

deploy/StableJoeStaking.js
cryptofish7: On the upgrade though, it will be "newly deployed" but it won't call initialize.

0x0Louis: Oh indeed, my bad I thought it was setting it on the implementation again

contracts/StableJoeStaking.sol
```diff
@@ -66,6 +67,9 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
     /// @dev Info of each user that stakes JOE
     mapping(address => UserInfo) private userInfo;
 
+    /// @dev Smol Joes contract address
+    IERC721Upgradeable smolJoes;
```
jummy123, line 71: NIT: Mark this variable explicitly as public (default if not specified) to match the style of other variables.

0x0Louis (approved):

jummy123:
LGTM

contracts/StableJoeStaking.sol
```diff
@@ -125,7 +135,11 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
     function deposit(uint256 _amount) external {
         UserInfo storage user = userInfo[_msgSender()];
 
-        uint256 _fee = _amount.mul(depositFeePercent).div(DEPOSIT_FEE_PERCENT_PRECISION);
+        uint256 _fee;
+        // Only EOAs holding Smol Joes are exempt from paying the deposit fee
+        if (address(smolJoes) != 0 && (smolJoes.balanceOf(_msgSender()) == 0 || _msgSender() != tx.origin)) {
```
jummy123, line 140: Tests are failing as you are comparing address with int. 0 needs to be cast to address `address(0)`

test/StableJoeStaking.test.js
```diff
@@ -672,6 +697,29 @@ describe("Stable Joe Staking", function () {
       );
     });
 
+    it("should allow EOAs with Smol Joes to be exempt from paying deposit fee", async function () {
+      await this.stableJoeStaking
+        .connect(this.dylan)
+        .deposit(ethers.utils.parseEther("100"));
```
jummy123, line 703: There doesn't look to be a bug there but for fuller test coverage you should also deposit from the vault before transferring the ERC721 to it.

contracts/StableJoeStaking.sol
jummy123: I think there is a potential for a problem here if you deploy this update and don't set the `smolJoe` variable immediately, as `&&` short circuits nobody would pay any fee on deposit until `smolJoe` address was set. It looks like this will be covered by the deployment script but still worth being aware.

jummy123: I think the following fixes the above problem.
```
if (!(address(smolJoes) != address(0) && smolJoes.balanceOf(_msgSender()) > 0 && _msgSender() == tx.origin)) {
     _fee = _amount.mul(depositFeePercent).div(DEPOSIT_FEE_PERCENT_PRECISION);
}
```

PR #120, opened 2023-09-19
Upgrade sjoe

0x0Louis:
The aim of this PR is to add a `sweep` function to sJoe for non-reward token (and not joe).
This PR also use immutable variables in order to save gas, to keep the storage ordering `__gap` variables are added.
A non-reentrant check was added for improved security.


contracts/StableJoeStaking.sol
```diff
@@ -40,28 +40,47 @@ contract StableJoeStaking is Initializable, OwnableUpgradeable {
...
     mapping(IERC20Upgradeable => uint256) public lastRewardBalance;
 
+    // @notice The address where deposit fees will be sent
     address public feeCollector;
-
     /// @notice The deposit fee, scaled to `DEPOSIT_FEE_PERCENT_PRECISION`
-    uint256 public depositFeePercent;
+    uint88 public depositFeePercent;
+    /// @notice Reentrancy guard
+    bool public reentrant;
```
Mathieu-Be, line 67: This only work because `depositFeePercent` is zero atm. That's a bit dangerous imo, imagine we put this upgrade on hold for some reason, then later we update the deposit fee and decide to finally carry on the upgrade while forgetting about this

Mathieu-Be (approved):

